PR #4, opened 2024-10-08
Nix Support

jaidetree:
This PR adds nix files to specify which tools are required to work on this project. This can then be combined with
[nix-direnv](https://github.com/nix-community/nix-direnv) by adding test `use flake` in a .envrc file and running `direnv allow .`. This uses nix to install the deps, and hooks into direnv so those binaries are readily available. Then all devs have to do is `cd jank-lang/clojure-test` and direnv will do the heavy lifting of linking the packages into the shell env so all that's needed is to run `lein test` and it works as expected.

This approach comes with several benefits:
- All devs are working with the same version via flake.lock.
- If you use flakes across other clojure projects you can have one version of clojure on disk, but at any point specify a specific version override on a project basis.
- Don't need to install these packages on the system. For example, you don't need to pollute your system with lein, it's only installed for this project.

To get started with nix, go through https://zero-to-nix.com/.

Summary:

- Adds a flake.nix with clojure, java, clj-kondo, and clojure-lsp
- Adds the flake.lock file

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1726560853,
+        "narHash": "sha256-X6rJYSESBVr3hBoH0WbKE5KvhPU5bloyZ2L4K60/fPQ=",
+        "owner": "numtide",
+        "repo": "flake-utils",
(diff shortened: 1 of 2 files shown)


jaidetree:
In theory this could be used for the CI too, then there would be a path for consistent packages between local & ci tasks.

jeaye:
Cool!

PR #6, opened 2024-10-18
Minor test additions

quoll:
This is just the function `aclone`. I've also added a common namespace for shared values in the numerical tests, and added a couple of tests for `bit-and`, `bit-or`, `bit-xor` and `bit-and-not`.

Nothing significant.

### DIFF
--- a/test/clojure/core_test/number_range.cljc
+++ b/test/clojure/core_test/number_range.cljc
@@ -1,0 +1,29 @@
+(ns clojure.core-test.number-range)
+
+(def MAX-INT #?(:clj Long/MAX_VALUE
+                :cljs js/Number.MAX_SAFE_INTEGER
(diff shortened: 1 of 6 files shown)


test/clojure/core_test/number_range.cljc
```diff
@@ -0,0 +1,29 @@
+(ns clojure.core-test.number-range)
+
+(def MAX-INT #?(:clj Long/MAX_VALUE
```
jeaye, line 3: Please lower case all of these vars. Sure, they're constants, but they're just vars.

quoll: I've adopted this. Your use of the word "constants" reminded me that this is a compiler hint that can be provided via metadata. I've added it, but am happy to remove.

PR #7, opened 2024-10-19
Tests for `*` and `*'`

quoll:
Minor additions

### DIFF
--- a/test/clojure/core_test/nan.cljc
+++ b/test/clojure/core_test/nan.cljc
@@ -1,0 +1,15 @@
+(ns clojure.core-test.nan
+  (:require [clojure.test :as t]))
+
+(t/deftest common
+  (t/is (thrown? Exception (NaN? nil)))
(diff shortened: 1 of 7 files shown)


test/clojure/core_test/nan.cljc
```diff
@@ -0,0 +1,15 @@
...
+
+(t/deftest common
+  (t/is (thrown? Exception (NaN? nil)))
+  (t/is (thrown? Exception (NaN? "##NaN")))
+  (t/are [in ex] (= ex (NaN? in))
+         0 false
+         1.0 false
+         -1.0 false
+         (Double/valueOf 1.0) false
+         (Double/valueOf 1.0) false
```
jeaye, line 12: Can we use `(double 1.0)` instead, since that doesn't rely on JVM interop?

quoll: Thank you, that's **much** better!

jeaye:
Thank you!

PR #8, opened 2024-10-29
'Or' tests

NoahTheDuke:
Plus some minor touches for `and`.

### DIFF
--- a/test/clojure/core_test/and.cljc
+++ b/test/clojure/core_test/and.cljc
@@ -3,10 +3,29 @@
 
 (t/deftest common
   (t/is (= true (and)))
+  (t/are [x] (= x (and x))
+    true
+    false
+    nil
+    :example)
   (t/are [ex a b] (= ex (and a b))
-         true true true
-         false true false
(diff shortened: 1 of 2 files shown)


test/clojure/core_test/and.cljc
```diff
@@ -3,10 +3,29 @@
...
+    true true true
+    nil true nil
+    nil nil true
+    nil nil nil)
+  (t/testing "binds values before comparing"
+    (let [counter (volatile! 0)]
+      (t/is (= nil (and (vswap! counter inc) nil)))
+      (t/is (= 1 @counter))))
+  (t/testing "early exits"
+    (t/is (= nil (and nil (last (range))))))
```
quoll, line 25: Is this a good idea? If the implementation has not implemented `and` correctly, then this will lock up.

NoahTheDuke: That's a good point. Better to have a failing test then a test that locks the runner.

jeaye:
Nice work!

PR #9, opened 2024-10-29
Tests for add-watch on atoms, refs, and agents

quoll:
I can't get watches on vars to respond, so I've had to leave them out.
EDIT: They're working now. I have no idea what I got wrong the last time I looked at this!


jeaye:
Thanks!

PR #10, opened 2024-10-30
Wrap bit-X NPE checks in reader conditionals

NoahTheDuke:
Javascript is incredibly permissive so it allows this naughtiness.


jeaye:
Thanks!

PR #12, opened 2024-11-30
Add tests for `nil?`

rob-3:


test/clojure/core_test/nil_questionmark.cljc
```diff
@@ -0,0 +1,16 @@
...
+    0 false
+    "" false
+    "nil" false
+    :nil false
+    ##Inf false
+    ##NaN false
+    0.0 false
+    {} false
+    '() false
+    [] false))
```
jeaye, line 16: It'd be good to add an infinite sequence test, á la: https://github.com/jank-lang/clojure.core-test/blob/main/test/clojure/core_test/any.cljc#L13

rob-3: Done. I can add it in a separate `deftest` if you prefer. I also added a test for `false` because I realized I missed it the first time around.

jeaye: Nice! I think it's good to be explicit about it in a separate test. `common` is meant for normal cases, while other edge cases have their own tests.

rob-3: Ok - changed!

jeaye: :muscle: Thanks for helping!

PR #14, opened 2024-12-01
Add tests for `inc` and `dec`

rob-3:
The ClojureScript tests I've tested manually in a ClojureScript REPL, but not with `lein test` since ClojureScript support hasn't been implemented yet.

### DIFF
--- a/test/clojure/core_test/dec.cljc
+++ b/test/clojure/core_test/dec.cljc
@@ -1,0 +1,17 @@
+(ns clojure.core-test.dec
+  (:require [clojure.test :as t]))
+
+(t/deftest common
+  (t/are [in ex] (= (dec in) ex)
+    1 0
+    14412 14411
+    -3 -4))
+
+(t/deftest underflow
+  #?(:clj (t/is (thrown? ArithmeticException (dec Long/MIN_VALUE)))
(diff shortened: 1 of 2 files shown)


test/clojure/core_test/dec.cljc
```diff
@@ -0,0 +1,17 @@
...
+    -3 -4))
+
+(t/deftest underflow
+  #?(:clj (t/is (thrown? ArithmeticException (dec Long/MIN_VALUE)))
+     :cljs (t/is (= (dec js/Number.MIN_SAFE_INTEGER) (- js/Number.MIN_SAFE_INTEGER 2)))))
+
+(t/deftest dec-nil
+  ;; ClojureScript says (= -1 (dec nil)) because JavaScript casts null to 0
+  #?(:clj (t/is (thrown? NullPointerException (dec #_:clj-kondo/ignore nil)))
+     :cljs (t/is (= -1 (dec #_:clj-kondo/ignore nil)))))
```
jeaye, line 17: I was just thinking as I did the dishes today, we didn't add a `nil` test for `even?` and `odd?`. Are you up for adding those?

rob-3: [Done](https://github.com/jank-lang/clojure.core-test/pull/16)

jeaye:
Nice work! Thanks for keeping this ball rolling. :grin:

PR #19, opened 2024-12-07
Add tests for rationals, `##Inf`, and `##NaN` to `star.cljc`

dgr:
Add tests for rationals, `##Inf`, and `##NaN` to `star.cljc`


jeaye:
Thanks!

PR #22, opened 2024-12-09
New tests for numeric and boolean predicates

dgr:
- `identical?`
- `zero?`
- `pos?`
- `neg?`
- `number?`
- `ratio?`
- `rational?`
- `integer?`
- `int?`
- `pos-int?`
- `nat-int?`
- `decimal?`
- `float?`
- `double?`
- `true?`
- `false?`
- `boolean`

### DIFF
--- a/test/clojure/core_test/boolean.cljc
+++ b/test/clojure/core_test/boolean.cljc
@@ -1,0 +1,56 @@
+(ns clojure.core-test.boolean
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.number-range :as r]))
+
+(deftest test-false?
+  (are [expected x] (= expected (boolean x))
+    true  0
+    true  1
+    true  -1
+    true  r/max-int
+    true  r/min-int
+    true  0.0
+    true  1.0
+    true  -1.0
+    true  (float 0.0)
+    true  (float 1.0)
+    true  (float -1.0)
+    true  (double 0.0)
+    true  (double 1.0)
+    true  (double -1.0)
+    true  r/max-double
+    true  r/min-double
+    true  ##Inf
+    true  ##-Inf
+    true  ##NaN
+    true  0N
+    true  1N
+    true  -1N
+    true  0/2
+    true  1/2
+    true  -1/2
+    true  0.0M
+    true  1.0M
+    true  -1.0M
+    false nil
+    true  true
+    false false
+    true  "a string"
+    true  "0"
+    true  "1"
+    true  "-1"
+    true  "true"
+    true  "false"
+    true  {:a :map}
+    true  #{:a-set}
+    true  [:a :vector]
+    true  '(:a :list)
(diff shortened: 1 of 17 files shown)


test/clojure/core_test/boolean.cljc
```diff
@@ -0,0 +1,56 @@
...
+    true  [:a :vector]
+    true  '(:a :list)
+    true  \0
+    true  \1
+    true  :a-keyword
+    true  :true
+    true  :false
+    true  :0
+    true  :1
+    true  :-1
```
NoahTheDuke, line 55: do you need 6 keywords? this seems to be testing a dialect's reader, not the function implementation, which should be covered by other tests.

NoahTheDuke: same with the rest of the tests

jeaye:
I like the addition of these additional keywords. Dave has good insight here into how a Clojure dialect might handle runtime objects. For example, in JS, we have implicit casts from "1" to 1. That sort of thing can be caught by this sort of test. At any rate, it's not harming anything.

NoahTheDuke:
No it's not, but if a dialect fails to parse `:true` correctly, do they want 30 identical failures, one for each usage?

You're right that this isn't hurting anything so i'll leave it alone lol. The minorest nitpick.

jeaye:
Woot! :tada: Nice work, Dave!

test/clojure/core_test/boolean.cljc
dgr: Not strictly. But cases like those are where weird things often come up. What I'm looking for is whether something is silently converted to something else in the platform, for whatever reason. Think of all the issues that JavaScript has had over the years with odd conversions. I'd note that `:0` and the other numeric ones aren't even strictly legal in Clojure (symbols and keywords are supposed to start with a non-numeric). But Clojure allows them, so it would be interesting to see what other implementations do. But sure, if you feel strongly about it, they aren't strictly necessary and can be removed.

dgr:
Put it this way... At this stage of the game, I'm like a thief trying to break into a building. I'm trying every doorknob I can find and twisting them to see if they are unlocked. Yes, they should all be locked and all my searching should be fruitless. But I've already found a couple of oddities with Clojure's implementation over the last couple of days in writing these tests. So, I'm sort of over-testing right now, even things that "should work" or seem redundant. My theory is that if running time or whatnot actually becomes and issue over time, we can pare things back. I'm sure we'll do that anyway as implementations start to run the suite, fail tests, and then say, "Hey, this test case is overly tight and we need to loosen things up." I'm fine with that. I'd rather have more tests at this point than less, even if some are redundant. As an aside, it's unlikely any of these tests extend running time much. They all are just a single function call. The cost of loading each test file, identifying the tests, etc., is way more work than these.

NoahTheDuke:
That makes a ton of sense, thanks!

PR #24, opened 2024-12-10
Add tests for various string output functions

dgr:
Add tests for

- `char`
- `char?`
- `format`
- `pr-str`
- `print-str`
- `println-str`
- `prn-str`
- `str`
- `string?`
- `subs`
- `with-out-str`


jeaye:
Woot! :tada:

PR #25, opened 2024-12-19
Add tests for various numeric casts

dgr:
Add tests for:

- `byte`
- `short`
- `int`
- `long`
- `float`
- `double`
- `bigint`
- `bigdec`
- `num`
- `rationalize`


jeaye:
Nice work! :muscle:

PR #26, opened 2024-12-21
Add tests for seq

angusiguess:
Tests covering `seq`

### DIFF
--- a/test/clojure/core_test/seq.cljc
+++ b/test/clojure/core_test/seq.cljc
@@ -1,0 +1,23 @@
+(ns clojure.core-test.seq
+  (:require [clojure.test :as t :refer [deftest testing is are]]))
+
+(deftest test-seq
+  ;; Sourced via canSeq https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/RT.java#L581
+  (are [in expected] (= expected (seq in))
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/seq.cljc
```diff
@@ -0,0 +1,30 @@
...
+                     :c 800
+                     nil 40}
+          input-sorted-map (into (sorted-map) input-map)
+          input-hash-map (into (hash-map) input-map)]
+      (is (= input (into #{} (seq input))))
+      (is (= input-hash (into (hash-set) (seq input))))
+      (is (= input-sorted-map (into (sorted-map) (seq input-sorted-map))))
+      (is (= input-hash-map (into (hash-map) (seq input-hash-map))))
+      (is (= input-map (into {} (seq input-map))))))
+  (testing "nonseqables"
```
jeaye, line 28: Do you mind adding a test case for infinite seqs? We want to make sure they're handled properly. Like we have here: https://github.com/jank-lang/clojure.core-test/blob/main/test/clojure/core_test/any.cljc#L13

angusiguess: sure thing!

jeaye:
Great job!

PR #27, opened 2024-12-21
Add tests for - and /

dgr:
Add new tests for:

- `-`
- `/`

### DIFF
--- a/test/clojure/core_test/minus.cljc
+++ b/test/clojure/core_test/minus.cljc
@@ -1,5 +1,6 @@
 (ns clojure.core-test.minus
-  (:require [clojure.test :as t :refer [deftest testing is are]]))
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.number-range :as r]))
 
 (deftest common
   (are [expected x y] (= expected (- x y))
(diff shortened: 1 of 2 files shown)


test/clojure/core_test/minus.cljc
```diff
@@ -0,0 +1,184 @@
...
+    ##-Inf ##NaN
+    ##NaN  ##NaN)
+
+  ;; Single arg
+  (is (= ##-Inf (- ##Inf)))
+  (is (= ##Inf (- ##-Inf)))
+  (is (NaN? (- ##NaN)))
+
+  (is (thrown? Exception (- ##NaN nil)))
+  (is (thrown? Exception (- ##Inf nil))))
```
jeaye, line 184: For both of these, it'd be good to have a test case for the variadic arity. i.e. `(apply - (range 20))`

dgr: Good point. I'll add.

dgr:
Sorry about the wait. I just pushed some updates that add more multi-arg and single-arg tests.

jeaye:
Excellent work! Thanks for all of the help. :slightly_smiling_face:

PR #28, opened 2024-12-23
Add test for compare

jake-jake-jake:
This adds tests for numeric, lexical, and some collection types for `compare`.


jeaye:
Nice work!

PR #30, opened 2024-12-31
Add test for partial

jake-jake-jake:


test/clojure/core_test/partial.cljc
```diff
@@ -0,0 +1,15 @@
...
+
+(defn test-fn [& args]
+  (into [] args))
+
+(t/deftest test-partial []
+  (let [simple-use (partial inc 2)]
+    (t/is (= 3 (simple-use))))
+  (let [lazily-evaluated (partial inc 2 3)]
+    #?(:clj  (t/is (thrown? Exception (lazily-evaluated)))
+       :cljs (t/is (thrown? js/Error  (lazily-evaluated)))))
```
jake-jake-jake, line 12: I don't know if handling the JS case is needed, but I thought I'd add this. I can remove if that makes more sense.

test/clojure/core_test/partial.cljc
```diff
@@ -0,0 +1,15 @@
+(ns clojure.core-test.partial
+  (:require [clojure.test :as t]))
+
+(defn test-fn [& args]
+  (into [] args))
+
+(t/deftest test-partial []
```
jeaye, line 7: What do you think about:

1. Chaining partials
2. Including lazy sequences
3. A partial partial

jake-jake-jake: Happy to add some more cases.

For 2, I am pushing up a test that has an infinite sequence and takes from it. If that's not quite right let me know.

For 1, do you mean a sequence of partials? I included what could be a test for that, but also, if that's not quite right let me know.

jeaye: 1. You got this one
2. Looks good
3. I was thinking something like `(((partial partial) +) 0 5)`

jake-jake-jake: Totally. I adjusted the `partial-partial` tests. Lemme know any other adjustments.

jeaye:
Well done!

PR #31, opened 2025-01-04
Add cases for zero arg, one arg, and multi-arg to arithmetic fns

dgr:
Adds some test cases for zero arg, single arg, and multi-arg to arithmetic functions (`+`, `-`, `*`, and `/`).


jeaye:
Nice!

PR #33, opened 2025-01-08
Add test cases for binding

jake-jake-jake:
I can add some more test cases for these if it makes sense to do so; these are the more obvious ones that occurred to me.

### DIFF
--- a/test/clojure/core_test/binding.cljc
+++ b/test/clojure/core_test/binding.cljc
@@ -1,13 +1,15 @@
 (ns clojure.core-test.binding
-  (:require [clojure.test :as t]))
+  (:require [clojure.test :as t]
+            [clojure.core-test.portability :as p]))
 
-(def  ^:dynamic *x* :unset)
-(def  ^:dynamic *y* nil)
-(defn ^:dynamic *f* [x] (inc x))
+(p/when-var-exists clojure.core/binding
+  (def  ^:dynamic *x* :unset)
+  (def  ^:dynamic *y* nil)
+  (defn ^:dynamic *f* [x] (inc x))
 
-(defn test-fn [] *x*)
+  (defn test-fn [] *x*)
 
-(t/deftest ^:heavy test-binding
+  (t/deftest ^:heavy test-binding
   ;; base-case with no overrides
   (t/is (= *x* :unset) "Unset is :unset")
   (t/is (= (*f* 1) 2)  "fn call")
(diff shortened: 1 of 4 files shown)


jeaye:
Great tests. What do you think about adding some tests for binding conveyance across threads? When a thread is created, it should copy the current binding state and changes made to bindings should only be visible on the current thread.

jake-jake-jake:
> Great tests. What do you think about adding some tests for binding conveyance across threads? When a thread is created, it should copy the current binding state and changes made to bindings should only be visible on the current thread.

Yeap, happy to add those. I wasn't sure we wanted to mess with threads in these, but if we are, we should test those cases.

I put up some that test threaded bindings via futures as well as how delay handles them. The tests are failing in CI, and I'll dig into a possible solution later this weekend/early next week, unless we already have a workaround for threaded execution. A quick search turns [this up](https://ask.clojure.org/index.php/10850/exception-using-futures-with-x-on-clojure-cli-1-10-3-929), which I think might be the cause (since the tests pass in my CIDER session).

jake-jake-jake:
I pushed up a hack to make this pass. It's not the most satisfying solution, but it does make things work for now. If there is a better way to direct lein to run specific tests alone, I am happy to change this to do that but I didn't see anything obvious in the docs or through digging around.

jeaye:
Nice! We just merged this PR: https://github.com/jank-lang/clojure.core-test/pull/37

Would you mind wrapping your new `binding` file in the same macro before I merge?

jake-jake-jake:
Done!

test/clojure/core_test/binding.cljc
```diff
@@ -0,0 +1,68 @@
...
+        (t/is (= @f :outer) "Thread context preserves binding context."))))
+  (binding [*x* :caller]
+    (let [f (future
+              (binding [*x* :callee]
+                (future (test-fn))))]
+      (binding [*x* :derefer]
+        (let [derefed-f @f]
+          (t/is (= :callee @derefed-f) "Binding in futures preserved.")))))
+
+  (shutdown-agents)))
```
jeaye, line 68: Hm, this will cause issues with any other tests using threads. Does it make sense to add a shutdown hook instead?

```clojure
; In portability.clj
(defn shutdown-agents-on-close []
  #?(:clj (.addShutdownHook (Runtime/getRuntime) (Thread. ^Runnable shutdown-agents))))
```

Taken from here: https://medium.com/helpshift-engineering/achieving-graceful-restarts-of-clojure-services-b3a3b9c1d60d

jake-jake-jake: It looks like that was what was happening with the threads in these tests! There was a different shutdown agents call in `add_watch.cljc`. I removed it and these tests pass without any issues now.

jeaye:
Great work!

PR #34, opened 2025-01-10
Add tests for `quot`, `rem`, and `mod`

dgr:
Add tests for:

- `quot`
- `rem`
- `mod`

Also adds a file called `portability.cljc` in which we can add various portability functions.

### DIFF
--- a/test/clojure/core_test/portability.cljc
+++ b/test/clojure/core_test/portability.cljc
@@ -1,0 +1,5 @@
+(ns clojure.core-test.portability)
+
+(defn big-int? [n]
+  (and (integer? n)
+       (not (int? n))))
--- a/test/clojure/core_test/mod.cljc
+++ b/test/clojure/core_test/mod.cljc
@@ -1,0 +1,85 @@
+(ns clojure.core-test.mod
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability :as p]))
+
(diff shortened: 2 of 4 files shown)


test/clojure/core_test/portability.cljc
```diff
@@ -0,0 +1,5 @@
+(ns clojure.core-test.portability)
+
+(defn big-int? [n]
+  (and (integer? n)
+       (not (int? n))))
```
jeaye, line 5: We probably want a reader conditional here, right? If CLJ, use `instance?`, otherwise, use the generic check?

dgr: We could, but that makes it very specific to JVM. The doc strings say that `int?` returns `true` for a fixed-width integer, whereas `integer?` returns true for all integers, whether fixed with or big. So, I figured this would be slightly more portable than a direct class check using `instance?`.

jeaye: Yep, I hear you. I was just wondering about the odds that there's an integer among Clojure dialects which isn't `int?` and which isn't a big int. :thinking: Maybe not worth considering.

dgr: I thought about that and decided we could cross that bridge later. It's easy enough to rewrite at that time.

jeaye:
Nice work! :muscle:

PR #38, opened 2025-01-15
Add tests for `rand` and `rand-int`

dgr:
Added tests for:

- `rand`
- `rand-int`

### DIFF
--- a/test/clojure/core_test/rand.cljc
+++ b/test/clojure/core_test/rand.cljc
@@ -1,0 +1,23 @@
+(ns clojure.core-test.rand
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability :as p]))
+
+(p/when-var-exists clojure.core/rand
+  (deftest test-rand
+    ;; Generally, we test that the numbers returned pass `double?` and
+    ;; that they are unique. Note that in theory `rand` could return
+    ;; the same double in the first 100 attempts, but that is highly
+    ;; unlikely.
+    (let [length 100]
+      (testing "zero-arg case"
+        (let [x (repeatedly length rand)]
+          (is (every? double? x))
+          (is (every? pos? x))
+          (is (apply distinct? x))))
+      (testing "one-arg case"
+        (let [limit 0.01 ; Choose something < 1 to constrain it further
+              x (repeatedly length #(rand limit))]
+          (is (every? double? x))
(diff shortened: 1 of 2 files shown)


test/clojure/core_test/rand_int.cljc
```diff
@@ -0,0 +1,17 @@
+(ns clojure.core-test.rand-int
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability :as p]))
+
+(p/when-var-exists clojure.core/rand-int
+  (deftest test-rand-int
+    ;; Generally, we test that the numbers returned pass `double?` and
```
jeaye, line 7: `double?` here can be `int?`.

test/clojure/core_test/rand.cljc
```diff
@@ -0,0 +1,23 @@
...
+        (let [x (repeatedly length rand)]
+          (is (every? double? x))
+          (is (every? pos? x))
+          (is (apply distinct? x))))
+      (testing "one-arg case"
+        (let [limit 0.01 ; Choose something < 1 to constrain it further
+              x (repeatedly length #(rand limit))]
+          (is (every? double? x))
+          (is (every? pos? x))
+          (is (apply distinct? x))
```
jeaye, line 22: I think that the `distinct?` check is asking for flakiness. Especially for ints. I'd rather have 0% flakiness and slightly less coverage than have >0% flakiness and slightly more coverage.

test/clojure/core_test/rand_int.cljc
dgr: Thanks. Copy/paste error on my part.

test/clojure/core_test/rand.cljc
dgr: Hm. OK. I could perhaps test them to make sure they are at least not constant. I feel like I need some test to ensure that they are changing. And to be clear, if there is a PRNG backing `rand`/`rand-int`, as opposed to a true RNG, the probability is pretty much zero that you'd have any duplicates. And I could reduce the probability even further by only testing 10 numbers, as opposed to 100. Or 5. Or 2.

dgr: OK, I just pushed a commit that changes the test to testing for non-constantness, which is a much lower bar.

jeaye:
Thanks, Dave!

PR #39, opened 2025-01-17
Make tests in `compare` more resilient and fix CLJS crash

dgr:
This makes the tests in `compare` more resilient by testing the result of compare using a predicate rather than equality with a specific number. Different implementations are free to return numbers of different magnitudes as long as the sign of the returned value is positive, negative, or zero.

This also fixes a CLJS compiler crash that occurs when comparing an infinite seq created with `(range)`.


jeaye:
Thanks!

PR #42, opened 2025-01-18
Add test for bound-fn

jake-jake-jake:
This adds tests for `bound-fn`. The helper `when-var-exists` seems not to work for macros. Do we have a solution for those or should I wrap the test in a reader macro for now?

### DIFF
--- a/test/clojure/core_test/binding.cljc
+++ b/test/clojure/core_test/binding.cljc
@@ -49,7 +49,6 @@
     (binding [*x* :here]
       (t/is (= @f :here) "Delayed functions inherit there bindings when forced"))
     (t/is (= @f :here) "And value persists outside binding expression"))
-  (Thread/sleep 1)
   (let [f (future (test-fn))]
     (binding [*x* :now-here]
       (t/is (= @f :unset) "Thread context is separate from joining thread")))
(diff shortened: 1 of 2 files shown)


test/clojure/core_test/binding.cljc
```diff
@@ -49,7 +49,6 @@
     (binding [*x* :here]
       (t/is (= @f :here) "Delayed functions inherit there bindings when forced"))
     (t/is (= @f :here) "And value persists outside binding expression"))
-  (Thread/sleep 1)
```
jake-jake-jake, line 52: I forgot to remove this when I was messing with `shutdown-agents` when writing the binding tests.

test/clojure/core_test/bound_fn.cljc
```diff
@@ -0,0 +1,49 @@
+(ns clojure.core-test.bound-fn
+  (:require [clojure.test :as t]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
+
+(def ^:dynamic *x* :unset)
+
+(t/deftest test-bound-fn
+  ;; base-case
```
jeaye, line 8: Please follow the other test files to use:

1. `when-var-exists` to wrap the one `deftest`
2. `testing` for each grouping (instead of comments)

jake-jake-jake: Updated!

jeaye:
Nice work!

PR #43, opened 2025-01-18
Convert `thrown?` clauses to handle CLJS and Clojure JVM

dgr:
This converts `(is (thrown? Exception ...))` clauses to `(is (thrown? #?(:cljs :default :default Exception) ...))` to properly catch any sort of Javascript exception in Clojurescript.


test/clojure/core_test/zero_questionmark.cljc
```diff
@@ -32,8 +32,8 @@
                      false -1.0M)
 
                    (is #?@(:cljs [(= false (zero? nil))]
-                           :default [(thrown? Exception (zero? nil))]))
+                           :default [(thrown? #?(:cljs :default :default Exception) (zero? nil))]))
```
NoahTheDuke, line 35: needed?

jeaye: Nope, but this conversion was automated. I saw it during the review, as well, but didn't think it was worth more iteration.

dgr: Yea, I'm going through and catching some of those by hand. I'll try to do a deeper, file-by-file pass.

PR #48, opened 2025-01-22
Clean up tests for CLJS `abs` to `intern`

dgr:
This covers tests from `abs` to `intern`, alphabetically.


jeaye:
Thanks, Dave! CLJS tests are getting there. :muscle:

PR #49, opened 2025-01-22
Clean up tests for CLJS from `keyword` to `symbol?`

dgr:
Fixes tests for CLJS from `keyword` to `symbol?`.


jeaye:
Nice! Only 6 failures remain!

PR #53, opened 2025-01-23
Add test for bound-fn*

jake-jake-jake:
This one is a little silly since it is basically the same as `bound-fn`, but here it is.


jeaye:
Nice! Really feels like `bound-fn*` should've been private, in `clojure.core`. :grin:

PR #54, opened 2025-01-24
Update taps and subs tests for clojurescript

jaidetree:
- Refactored clojure tap test to make it more reliable
- Draft alternative, clojurescript compatible version of the test
- Fix subs test in clojurescript

### DIFF
--- a/test/clojure/core_test/subs.cljc
+++ b/test/clojure/core_test/subs.cljc
@@ -1,5 +1,5 @@
 (ns clojure.core-test.subs
-  (:require [clojure.test :as t :refer [deftest testing is are]]
+  (:require [clojure.test :as t :refer [deftest is]]
             [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
 
 (when-var-exists clojure.core/subs
@@ -17,7 +17,10 @@
          (is (= "abcde" (subs "abcde" -1)))
          (is (= "abc" (subs "abcde" -1 3)))
          (is (= "" (subs "abcde" -1 -3)))
-         (is (thrown? :default (subs nil 1 2)))
+         (try
+           (subs nil 1 2)
+           (catch :default err
+             (is (= (.-message err) "Cannot read properties of null (reading 'substring')"))))
          (is (= "ab" (subs "abcde" nil 2)))
          (is (= "a" (subs "abcde" 1 nil)))]
         :default
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ npx shadow-cljs watch test
 If you only want to autorun specific test files you may run the following:
 
 ```bash
-npx shadow-cljs watch app --config-merge '{:autorun false}'
+npx shadow-cljs watch test --config-merge '{:autorun false}'
 ```
 
 In another terminal, run the following, multiple namespaces are comma (,)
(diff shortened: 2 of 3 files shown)


test/clojure/core_test/subs.cljc
```diff
@@ -17,7 +17,10 @@
          (is (= "abcde" (subs "abcde" -1)))
          (is (= "abc" (subs "abcde" -1 3)))
          (is (= "" (subs "abcde" -1 -3)))
-         (is (thrown? :default (subs nil 1 2)))
+         (try
+           (subs nil 1 2)
+           (catch :default err
+             (is (= (.-message err) "Cannot read properties of null (reading 'substring')"))))
```
dgr, line 23: What is the point of this change to `subs`?

jaidetree: image

```
FAIL in (test-subs) (clojure/core_test/subs.cljc:20:14)
expected: (thrown? :default (subs nil 1 2))
  actual: nil
```

It was the only test left that was failing. Tried replacing `:default` with `js/Error` and `js/TypeError` but it was still failing. These changes seem to be working consistently.

dgr: Hm. Odd. I don't see that on my system. We should try to understand why that is and why `(thrown? :default ...)` doesn't work. If it throws, then `:default` should catch it, right? It should catch everything since it's effectively the same as catching `Throwable` in JVM.

dgr: If I execute this in a cljs REPL, I get:
```
cljs.user=> (subs nil 1 2)
Execution error (TypeError) at (<cljs repl>:1).
Cannot read properties of null (reading 'substring')
```

dgr: @jaidetree , what OS are you using? I'm running on MacOS 15.2 (Sequoia with latest updates applied).

jaidetree: That's error that is encountered in ci too

jaidetree: My thinking was to get the fix merged first, then investigate further. Every push since those changes have had fully passing ci which is a pretty big milestone. It's not like it's worse test code, just not as concise.

Running Mac OS 15.2 as well and same cljs:
```
cljs.user=> *clojurescript-version*
"1.11.132"
```

jaidetree: It was passing on my local too but was always failing in ci

jaidetree: Did find one factor here: The test passes when running `npx shadow-cljs watch test` but fails if you run:

```sh
npm run build
node target/js/node-tests.js
```

This feels out of scope for this PR to fix that

dgr: > My thinking was to get the fix merged first, then investigate further. Every push since those changes have had fully passing ci which is a pretty big milestone. It's not like it's worse test code, just not as concise.
> 
> Running Mac OS 15.2 as well and same cljs:
> 
> ```
> cljs.user=> *clojurescript-version*
> "1.11.132"
> ```

Yea, to be clear, I'm OK with merging the PR and there is benefit to having all the tests pass CI. I just want us to understand why the issue is there so that it doesn't get lost. In particular, testing for thrown exceptions is pretty basic, so it needs to work. If we're jumping through hoops to get something that will pass and we're not understanding why, that concerns me. And in particular, if we can run a test locally and have it pass and then it fails in CI, we need to understand that. It makes it difficult to create tests, otherwise.

PR #58, opened 2025-01-26
Fix `add-watch` to test watching atoms in cljs

dgr:
Also splits out tests for watching vars to a separate conditional section since cljs doesn't support watching vars.


jeaye:
Thanks, Dave!

PR #60, opened 2025-01-27
Add tests for variations of `drop` and `take`

dgr:
Add tests for
- `drop`
- `drop-last`
- `drop-while`
- `take`
- `take-last`
- `take-while`

### DIFF
--- a/test/clojure/core_test/drop.cljc
+++ b/test/clojure/core_test/drop.cljc
@@ -6,5 +6,23 @@
   (deftest test-drop
     (is (= (range 1 10) (drop 1 (range 0 10))))
     (is (= (range 5 10) (drop 5 (range 0 10))))
+    (is (= 5 (first (drop 5 (range))))) ; lazy version handles infinite `range`
+    (is (= '() (drop 5 nil)))           ; nil acts as empty list
+
     ;; Transducer version
-    (is (= (vec (range 5 10)) (into [] (drop 5) (range 0 10))))))
+    (is (= (vec (range 5 10)) (into [] (drop 5) (range 0 10))))
+
+    ;; Note that we can drop from other types of collections, but
+    ;; because they are not sequential, we don't know exactly what the
+    ;; result will be. This tests that something truthy remained after
+    ;; dropping one item and that `drop` didn't throw when given maps
+    ;; or sets. We need `doall` here to force the realization of the
+    ;; lazy seq created by `drop`.
+    (is (doall (drop 1 {:a 1 :b 2 :c 3})))
+    (is (doall (drop 1 #{:a :b :c})))
+
+    ;; Negative tests
+    (is (thrown? #?(:cljs :default, :default Exception)
+                 (doall (drop nil (range 0 10)))))
+    (is (thrown? #?(:cljs :default, :default Exception)
+                 (into [] (drop nil) (range 0 10))))))
--- a/test/clojure/core_test/drop.cljc
+++ b/test/clojure/core_test/drop.cljc
@@ -1,0 +1,10 @@
+(ns clojure.core-test.drop
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
+
+(when-var-exists clojure.core/drop
+  (deftest test-drop
+    (is (= (range 1 10) (drop 1 (range 0 10))))
+    (is (= (range 5 10) (drop 5 (range 0 10))))
+    ;; Transducer version
+    (is (= (vec (range 5 10)) (into [] (drop 5) (range 0 10))))))
--- a/test/clojure/core_test/drop_last.cljc
+++ b/test/clojure/core_test/drop_last.cljc
@@ -4,7 +4,16 @@
 
 (when-var-exists clojure.core/drop-last
   (deftest test-drop-last
-    ;; by default, drop the last item
+    ;; drop the last item
     (= (range 0 9) (drop-last (range 0 10)))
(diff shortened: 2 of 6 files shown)


test/clojure/core_test/drop.cljc
```diff
@@ -0,0 +1,10 @@
+(ns clojure.core-test.drop
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
+
+(when-var-exists clojure.core/drop
+  (deftest test-drop
+    (is (= (range 1 10) (drop 1 (range 0 10))))
+    (is (= (range 5 10) (drop 5 (range 0 10))))
+    ;; Transducer version
+    (is (= (vec (range 5 10)) (into [] (drop 5) (range 0 10))))))
```
jeaye, line 10: Makes sense to add a test for those which support infinite seqs (drop, drop-while, take, take-while).

test/clojure/core_test/drop.cljc
```diff
@@ -0,0 +1,10 @@
+(ns clojure.core-test.drop
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
+
+(when-var-exists clojure.core/drop
```
jeaye, line 5: All of these accept `nil` input, too, right? Anything else throws?

test/clojure/core_test/drop.cljc
dgr: Yep

test/clojure/core_test/drop.cljc
dgr: Yes, will add.

test/clojure/core_test/drop.cljc
dgr: Added infinite lazy seq tests

test/clojure/core_test/drop.cljc
dgr: Added `nil` testing

jeaye:
> Hm. This is failing cljs CI test with:
> 
> ```
> Testing clojure.core-test.take
> 
> FAIL in (test-take) (clojure/core_test/take.cljc:17:9)
> expected: (thrown? :default (doall (take nil (range 0 10))))
>   actual: nil
> 
> FAIL in (test-take) (clojure/core_test/take.cljc:19:9)
> expected: (thrown? :default (into [] (take nil) (range 0 10)))
>   actual: nil
> 
> Testing clojure.core-test.take-last
> 
> FAIL in (test-take-last) (clojure/core_test/take_last.cljc:14:9)
> expected: (thrown? :default (doall (take-last nil (range 0 10))))
>   actual: nil
> ```
> 
> I do NOT see this in cljs testing on my local machine and I'm not quite sure what is happening. @jaidetree, any ideas on this? It's looking like some issue of laziness, which I was dealing with earlier this morning, but the wrapping `doall` should have fixed that, and `into` is eager.

JS (and thus CLJS) is crazy. Is `nil` being punned into a 0 here, once we compile with `:simple`? Probably best way to test, at this point is a `println`, unless we can replicate the package versions that CI is using.

dgr:
I was wondering about `nil` punning to zero as well. I have seen that in lots of other tests, but it's not clear to me in which cases it is done and which it isn't. Specifically, I'm trying to understand why the tests pass in one environment and fail in another. Unfortunately, I just don't know enough about the CLJS compiler, optimizations, and compilation phases to figure it out. One goal we should all have, however, is that the local test running environment should be, as much as is possible, identical in behavior to the CI environment. If a test passes locally, it should pass in CI for a PR. If it fails locally, I should be able to debug it locally.

jaidetree:
> > Hm. This is failing cljs CI test with:
> > ```
> > Testing clojure.core-test.take
> > 
> > FAIL in (test-take) (clojure/core_test/take.cljc:17:9)
> > expected: (thrown? :default (doall (take nil (range 0 10))))
> >   actual: nil
> > 
> > FAIL in (test-take) (clojure/core_test/take.cljc:19:9)
> > expected: (thrown? :default (into [] (take nil) (range 0 10)))
> >   actual: nil
> > 
> > Testing clojure.core-test.take-last
> > 
> > FAIL in (test-take-last) (clojure/core_test/take_last.cljc:14:9)
> > expected: (thrown? :default (doall (take-last nil (range 0 10))))
> >   actual: nil
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I do NOT see this in cljs testing on my local machine and I'm not quite sure what is happening. @jaidetree, any ideas on this? It's looking like some issue of laziness, which I was dealing with earlier this morning, but the wrapping `doall` should have fixed that, and `into` is eager.
> 
> JS (and thus CLJS) is crazy. Is `nil` being punned into a 0 here, once we compile with `:simple`? Probably best way to test, at this point is a `println`, unless we can replicate the package versions that CI is using.

Oddly when I was doing `(println (subs nil 1 2))` it was not getting factored out to nil by the google closure compiler. That said, since migrating to `:simple` from `:advanced` it's not likely to be caused by a compiler optimization but I'll poke around the compiled code.

jaidetree:
Fixed in #61, the issue was :elide-asserts was set to true by default on release builds. The tests were all passing with that set to `false`, hopefully there's not too many extra config options like that to fix.

dgr:
Thanks, @jaidetree ! @jeaye , I guess you can apply both this PR as well as #61 .

jeaye:
Nice work!

PR #61, opened 2025-01-28
Disable elide-asserts in release builds

jaidetree:
Shadow-cljs defaults to eliding assert calls in release builds but not development builds. This makes behavior more consistent between dev and releases.

Also of note, the tests were compiled with elide-asserts false and optimizations advanced to see if we still need to use simple. The results were that subs failed again so for now optimizations should be simple and elide-asserts need to be false.


jaidetree:
Resolves issue found in #60 where tests fail inexplicably.

jeaye:
Great find!

PR #65, opened 2025-01-29
Add tests for zipmap

jake-jake-jake:
If there is an easy way to check equality of anonymous functions, infinite ranges, I can add a few tests with those as elements of the consumed seqs, but I don't know how to do that simply. These are the cases that came to mind. Happy to add others if any seem obvious.


jeaye:
Thank you! Looks good. Just pending the CLA signoff.

jake-jake-jake:
Git config issues. Should be good now!

jeaye:
Nice work!

PR #67, opened 2025-01-31
Add tests for `sort`

bpiel:


test/clojure/core_test/sort.cljc
```diff
@@ -0,0 +1,20 @@
+(ns clojure.core-test.sort
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
+
+(when-var-exists
+ clojure.core/sort
```
jeaye, line 6: Clojure (the JVM) specifically calls out that this is a stable sort. We took steps in jank to replicate this. We should test for this case, too. Doing so will require a bit of planning, since we'll want to be sure the objects haven't been reordered. Just a simple `=` won't do that, so I recommend putting that into a `testing` and then wrapping the existing `are` in a `testing` for the common cases.

Let me know if you want some guidance here.

test/clojure/core_test/sort.cljc
```diff
@@ -0,0 +1,20 @@
...
+    (are [expected x] (= expected (sort x))
+      '() nil
+      '(1 2 3) [3 1 2]
+      '(nil 2 3) [3 nil 2]
+      '(1 2 3) '(3 1 2)
+      '(1 2 3) #{3 1 2}
+      '([:a 1] [:b 2]) {:b 2 :a 1}
+      '([1] [2] [3]) [[3] [1] [2]]
+      '("a" "b" "c") ["b" "a" "c"]
+      '(\c \e \j \l \o \r \u) (sort "clojure"))
```
jeaye, line 17: In this case, you're sorting twice. Is that intentional?

jeaye (changes requested):

PR #68, opened 2025-02-04
Add tests for `shuffle`

bpiel:


test/clojure/core_test/shuffle.cljc
```diff
@@ -0,0 +1,34 @@
+(ns clojure.core-test.shuffle
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
+
+(when-var-exists clojure.core/shuffle
+  (deftest test-shuffle
+    (testing "basic shuffle behaviour"
+      ;; yes, technically, this could produce a false positive,
+      ;; but it's *extremely* unlikely
```
jeaye, line 9: How can we test shuffle without this? For example, the `rand` tests: https://github.com/jank-lang/clojure.core-test/pull/38#discussion_r1917516830

Since it's based on randomness, unless we could fix the seed (which we can't, in a cross-platform way), we can't test it. I suspect the only thing we can do, in an airtight fashion, is the sort of tests you have below.

jeaye (changes requested):

test/clojure/core_test/shuffle.cljc
bpiel: I understand why you want to drop this section, and I'm fine with going that route, but I do want to clarify that, in this case, "*extremely*" means on the order of being able to run this every second til the end of time and not getting a false positive.

jeaye: I hear ya. Zero chance of failure is still preferable to non-zero. Losing some test coverage is acceptable to me.

jeaye:
Looks good!

PR #69, opened 2025-02-05
Add tests for interleave/interpose 

jake-jake-jake:


jake-jake-jake:
@jeaye I struggled to get the reader macro tests to actually pass. I am not sure why putting the reader conditionals in would result in a `nil` result from `thrown?` rather than properly identifying the illegal argument exception. I can add them back in if there is a known work around, or I can spend more time digging into them if it is needed.

jeaye:
> @jeaye I struggled to get tests for bad inputs wrapped in reader macros to actually pass. I am not sure why putting the reader conditionals in would result in a `nil` result from `thrown?` rather than properly identifying the illegal argument exception. I can add them back in if there is a known work around, or I can spend more time digging into them if it is needed.

Not sure what you mean here. Based on what I can tell from the CI failures in previous commits, you're using `function?` when you want to be using `fn?`. There is no `clojure.core/function?`.

jake-jake-jake:
> Not sure what you mean here. Based on what I can tell from the CI failures in previous commits, you're using `function?` when you want to be using `fn?`. There is no `clojure.core/function?`.

I readded the macros. It's strange behavior:

```
#?(:clj (is (thrown? Exception (interpose 1 :not-collection))))
```

The `thrown?` test returns nil rather than an exception from with the reader macro. I am not sure if that has to do with interpose/interleave using macros underneath the hood, or something else. The tests work fine when called directly, but then they would be not work for JS.

jake-jake-jake:
Actually it fails even without the reader macro. I am going to remove it.

jeaye:
Nice work!

PR #70, opened 2025-02-06
Some clojure.string tests

frenchy64:
Also included `subs`.


jeaye (changes requested):
Tests look great. bb changes, too. Can you please add some negative tests, and edge-case tests, though? i.e. what happens when:

* Input is nil
* Input is not a string
* Input is already capitalized

Try to break it and capture its behavior in tests.

jeaye:
Is this ready for review again? Trying to clean up the dozen PRs we have open.

frenchy64:
Yes I think so.

jeaye:
Looks good! Thanks, Ambrose.

PR #71, opened 2025-02-06
Prep jank support

frenchy64:
- add deps.edn
- `(thrown? Exception form)` => `(thrown? form)`


test/clojure/core_test/count.cljc
```diff
@@ -2,6 +2,9 @@
   (:require [clojure.test :as t :refer [deftest testing is are]]
             [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
 
+#?(:cljs nil
+   :default (def a-char \a))
+
```
jeaye, line 7: This is bleeding a jank bug into these tests. We can't do this. Instead, we need to fix the jank bug and have these tests fail until it's fixed.

jeaye (changes requested):

jeaye:
Nice work, Ambrose!

frenchy64:
Thanks Jeaye!

PR #72, opened 2025-02-10
Add most basic = tests

rob-3:
Wasn't sure how you want to handle math and other platform quirks/inconsistencies, but I at least got the tests passing for JVM Clojure and CLJS. There is also the question of testing `not=` - it's easy to just copy over the tests but I wonder if it might be nicer to do something like:
```clojure
(t/are [in ex] (and (= in ex) (not (not= in ex)))
  ;; ...
```
in terms of making sure those tests stay in sync.

### DIFF
--- a/test/clojure/core_test/eq.cljc
+++ b/test/clojure/core_test/eq.cljc
@@ -1,0 +1,136 @@
+(ns clojure.core-test.eq
+  (:require [clojure.test :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists]]))
+
+;; Doesn't test (yet):
+;; * Java collections for JVM Clojure
+;; * clojure.lang.PersistentQueue
+;; * ##Inf and some other special forms
+;; * Records
+
+(defn tests [eq]
+  (testing "two scalars"
+    (are [in ex] (eq in ex)
+      nil nil
+      true true
+      false false
+      \a \a
+      "green" "green"
+      :eggs :eggs
+      :and/ham :and/ham
+      'one-fish 'one-fish
+      'two/fish 'two/fish
+      42 42
+      3.14 3.14
+      3.141592M 3.141592M))
+
+  (testing "two scalars unequal"
+    (are [in ex] (not (eq in ex))
+      nil false
+      true false
+      \a \b
+      "yellow" "purple"
+      :hello :goodbye
+      :my/hello 'my/hello
+      :my/hello :your/hello
+      'one-fish 'two-fish
+      'red/fish 'red/coral))
+
+  (testing "a string is not a list of chars"
+    (is (not (eq "hello" '(\h \e \l \l \o)))))
+
+  (testing "collections"
+    (are [in ex] (eq in ex)
+      [] '()
+      '() []
+      [1 2 3] '(1 2 3)
+      '(1 2 3) [1 2 3]
+      [0 1 2] (range 3)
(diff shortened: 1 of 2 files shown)


test/clojure/core_test/eq.cljc
```diff
@@ -0,0 +1,116 @@
+(ns clojure.core-test.eq
+  (:require [clojure.test :as t]))
+
+;; Doesn't test (yet):
+;; * Java collections for JVM Clojure
+;; * clojure.lang.PersistentQueue
+;; * ##Inf and some other special forms
+;; * Records
+
+(t/deftest two-scalars
```
jeaye, line 10: Please follow the existing tests. Here's a reference: https://github.com/jank-lang/clojure-test-suite/blob/main/test/clojure/core_test/inc.cljc

1. Wrap everything in `when-var-exists`
2. Use one `deftest` with multiple `testing` forms within

You can accomplish both of these by using our babashka task for creating tests! Just run

```bash
bb new-test =
```

test/clojure/core_test/eq.cljc
```diff
@@ -0,0 +1,116 @@
...
+              r' r]
+          (= r r'))))
+
+(t/deftest functions
+  ;; identical? functions are =, but no other functions
+  (t/is (not (= #(+ 2 %) #(+ 2 %))))
+  (t/is (let [f #(+ 2 %)
+              f' f]
+          (= f f'))))
+
```
jeaye, line 100: `=` also has a variadic arity which we'll want to test.

test/clojure/core_test/eq.cljc
```diff
@@ -0,0 +1,116 @@
...
+    true true
+    false false
+    \a \a
+    "green" "green"
+    :eggs :eggs
+    :and/ham :and/ham
+    'one-fish 'one-fish
+    'two/fish 'two/fish
+    42 42
+    84/2 42
```
jeaye, line 22: `84/2` actually reads as `42`, not a ratio. If we're trying to compare a ratio to a number, we'll need something which reads as a ratio.

jeaye (changes requested):
Hi! Thanks for the PR. :slightly_smiling_face: These are some good tests. Please follow the comments to update things accordingly and then request my feedback again.

test/clojure/core_test/eq.cljc
rob-3: Good catch; I switched to 22/7 instead

test/clojure/core_test/eq.cljc
rob-3: I added some tests for this, lmk if something more comprehensive would be better.

test/clojure/core_test/eq.cljc
rob-3: I manually migrated my code over to this format.

jeaye:
These look great now. I think it's a good question, how to support `=` and `not=`. We don't want to test multiple vars in one file; each var gets its own. However, maybe what we do is to pull this `deftest` code into a `defn` which takes the fn. For the eq tests, we give it `=`. For the noteq tests, we give it `(complement not=)`. That would allow us to use the same tests for both. The noteq ns would just require the eq ns and call the fn in there as part of its own `deftest`.

rob-3:
> Would you like to do that as part of this PR?

It is done. I also learned about a strange `not=` inconsistency I never knew about with `##NaN`s on the JVM.

rob-3:
@dgr They're all yours! I am not working on them right now.

PR #73, opened 2025-02-19
Add tests for `case`

dgr:
Add tests for the `case` macro.

### DIFF
--- a/test/clojure/core_test/case.cljc
+++ b/test/clojure/core_test/case.cljc
@@ -24,8 +24,8 @@
       nil :nil-result
       ##Inf :inf-result
       ##-Inf :negative-inf-result
-      #?@(:clj ['(list of syms) :list-of-syms-result] ; Note that these are not alternatives in CLJ
-          :default [])
+      #?(:cljs ('(list of syms))        ; CLJS wants to eval the inner list, even though it shouldn't
+         :default ((list of syms))) :list-of-syms-result ; if you need to match a list, wrap it in another list
       [:vec :of :kws] :vec-of-kws-result
       {:a :map :of :kws} :map-of-kws-result
       #{:a :set :of :kws} :set-of-kws-result
@@ -63,7 +63,7 @@
     (case x
       (range 0 5) :bad-range-result ; range not eval'd. matches 'range, 0, or 5.
       (1 2 3 4) :real-range-result  ; can't include 0 because previous clause
-      'foo :quote-foo-result ; reader expands to (quote foo), a list of alternatives
+      'foo :quote-foo-result ; reader expands to `(quote foo)`, a list of alternatives
       ##NaN :nan-result
       ;; empty default clause, so will throw if no matches
       ))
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/case.cljc
```diff
@@ -0,0 +1,174 @@
...
+      #?@(:cljs []
+          :default
+          [1/2 :ratio-result])
+      \a :character-result              ; CLJS reader will convert \a to "a"
+      true :boolean-true-result
+      false :boolean-false-result
+      nil :nil-result
+      ##Inf :inf-result
+      ##-Inf :negative-inf-result
+      #?@(:clj ['(list of syms) :list-of-syms-result] ; Note that these are not alternatives in CLJ
```
NoahTheDuke, line 27: how does this not get read in as a list of alternatives?

test/clojure/core_test/case.cljc
```diff
@@ -0,0 +1,174 @@
...
+        0 :bad-range-result
+        1 :real-range-result
+        2 :real-range-result
+        3 :real-range-result
+        4 :real-range-result
+        5 :bad-range-result
+        'quote :quote-foo-result
+        'foo :quote-foo-result)
+
+      (is (thrown? #?(:cljs :default, :default Exception) (negative-tests ##NaN)))
```
jeaye, line 173: Will you please use `:cljs :default, :clj Exception` instead? We have jank running these tests now and `Exeption` is not a sane default, since it's CLJ specific.

test/clojure/core_test/case.cljc
```diff
@@ -0,0 +1,174 @@
...
+        #?@(:cljs ["a" :character-result]) ; since chars are single character strings
+        true :boolean-true-result
+        false :boolean-false-result
+        nil :nil-result
+        ##Inf :inf-result
+        ##-Inf :negative-inf-result
+        #?@(:clj ['(list of syms) :list-of-syms-result]) ; works in CLJ to match a list
+        [:vec :of :kws] :vec-of-kws-result
+        '(:vec :of :kws) :vec-of-kws-result ; `case` allows vec pattern to match a list
+        #?@(:clj ['[list of syms] :list-of-syms-result]) ; works in CLj
```
jeaye, line 124: If CLJS doesn't support this, let's do `:cljs [] :default ...` since other dialects should support this. Same goes for the other cases where CLJS doesn't work.

jeaye (changes requested):
Excellent tests, Dave! We have jank support in this repo now (happened while you were out), so I have some comments to help make that work.

test/clojure/core_test/case.cljc
dgr: I think that's exactly what's happening.  In the `case` form, it's expanding `'(list of syms)` to `(quote (list of syms))` (since the reader adds the `quote` and wraps the quoted item). Then, when you pass in `(list of syms)` to the `case` form, it matches the second item in that list. On thinking about it, I actually look for that case more explicitly in the other tests, so I'm going to remove it along with the changes @jeaye asked for.

test/clojure/core_test/case.cljc
dgr: OK, will do.

test/clojure/core_test/case.cljc
dgr: I'm going to remove this, per the comment to @NoahTheDuke. I think it's better tested in the negative tests.

test/clojure/core_test/case.cljc
dgr: OK, I fixed this.

test/clojure/core_test/case.cljc
dgr: Done.

test/clojure/core_test/case.cljc
dgr: Done.

jeaye:
Looks great, Dave. Thank you!

PR #78, opened 2025-02-20
Add tests for `vector?` and `list?`

dgr:
Add tests for:
* `vector?`
* `list?`

### DIFF
--- a/test/clojure/core_test/vector_qmark.cljc
+++ b/test/clojure/core_test/vector_qmark.cljc
@@ -6,6 +6,7 @@
   (deftest test-vector?
     (are [expected x] (= expected (vector? x))
       true [1 2 3]
+      true (first {:a 1})               ; map entries are also vectors
       false (sorted-map :a 1)
       false (sorted-set :a)
       false '(1 2 3)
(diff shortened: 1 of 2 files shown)


test/clojure/core_test/vector_qmark.cljc
```diff
@@ -0,0 +1,29 @@
+(ns clojure.core-test.vector-qmark
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
+
+(when-var-exists clojure.core/vector?
+  (deftest test-vector?
+    (are [expected x] (= expected (vector? x))
+      true [1 2 3]
```
jeaye, line 8: I think it'd be good to have a test for map entry here, since it acts like a vector.

jeaye (changes requested):

test/clojure/core_test/vector_qmark.cljc
dgr: Ah, good catch. I'll add that.

dgr: OK, this is done. Sorry for the delay.

jeaye (approved):
Thanks, Dave!

PR #81, opened 2025-03-08
Add tests for `reduce`

cddr:
Add tests for:

 * reduce

### DIFF
--- a/test/clojure/core_test/reduce.cljc
+++ b/test/clojure/core_test/reduce.cljc
@@ -33,7 +33,28 @@
                       :cljs js/Error) (reduce nil nil)))
       (is (= 6 (reduce + 0 [1 2 3]))))
 
-    (testing "edge cases"
+    (testing "val is not supplied"
+      (is (= 3 (reduce (fn [a b]
+                         (+ a b))
+                       [1 2])))
+
+      (testing "empty coll"
+        (is (= 1 (reduce (fn [] 1) []))))
+
+      (testing "coll with 1 item"
+        (is (= 1 (reduce (fn []
+                           (is false)
+                           (throw (ex-info "should not get here" {})))
+                         [1])))))
+
+    (testing "val is supplied, empty coll"
+      (is (= 1 (reduce (fn []
+                           (is false)
+                         (throw (ex-info "should not get here" {})))
+                       1
+                       []))))
+
+    (testing "reduction by type"
       (let [int-new (interop :int-new)
             char-new (interop :char-new)
             byte-new (interop :byte-new)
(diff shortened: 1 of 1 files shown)


frenchy64:
There's a few more edge cases documented in the `reduce` docstring if you're interested. Also `reduced`.

jeaye:
`reduced` is its own var, so that should go into its own file. @cddr are you interested in expanding on this for the edge cases which @frenchy64 mentioned?

jeaye:
Thanks! Have you added everything you'd like to add, at this point?

jeaye:
Sounds great. Thanks for the help!

dgr:
Hi @cddr, 

I was reviewing the tests in `reduce` a couple days ago and had some questions. It was not always obvious to me why the test code is as it is. If you could give me some background on the intent here, that would help me.

All line numbers correspond with git hash fc224d7, though pretty much any version since the PR was committed works, too.

1. What is the intent of the `:int-new` key in the `interop` map (line 8)? Why not just have this be a standalone function (e.g., `(defn init-new ...)`) that has conditional reader macros appropriately?
2. What is the purpose of the `:int-new` function (line 8)? Are you just trying to force boxing of primitive values? If so, why?
3. You return `Long/TYPE` in the `:Float` case of `interop` for Clojure JVM (line 31). Is that right? Do you really want to return `Float/TYPE`?
4. On line 76 you have a comment that says "enough to cross nodes". What does that mean, exactly? What is a "node?"
5. Why do you only call `int-new` for integers, not similar functions for long, float, double, etc. (line 80)?
6. Why the cast to `int` before calling `int-new` which then forces it to `Integer` in the case of JVM and then whatever the int types are for other dialects? I don't understand what this is trying to do and specifically what it's trying to validate.
7. What are you trying to test when you're creating the various types of low-level Java arrays? Given that most non-JVM dialects will not have the same underlying array types, this seems a bit overkill, but maybe I'm missing something. So, while I could see testing an array type (maybe a generic object array and perhaps an unboxed int array or something), the rest feels like more than we need. But perhaps I'm missing it.

Thanks for your help.

dgr:
@cddr , ping...

cddr:
Hi @dgr, good questions. This was a while ago and the code seems to have some additions since I was there. I must confess I used Claude to help me with this response.

1 & 2. The `interop` map just bundles all the dialect-specific reader conditionals in one place. It would be equally valid as standalone functions but this felt tidier. The purpose of `:int-new` is to construct a boxed `Integer` — you need it because Clojure numeric literals default to `Long`, but `int-array` needs `Integer` objects.

3. Yep, that's a bug. Should be `Float/TYPE`. Apologies this got through.  Not sure how.

4. Clojure's persistent vectors are 32-way trees internally. A "node" is one of those tree nodes. `(range 1 100)` gives 99 elements which spans multiple nodes, so we're testing that `reduce` correctly traverses the whole tree rather than just the first 32 elements.

5 & 6. The other typed arrays (`long-array`, `float-array`, `double-array`) accept Clojure's native literal types directly. `int-array` is the awkward one because Clojure's integer literals are `Long` by default, so you need the explicit `(int %)` cast to get the right primitive type.

7. The intent was to verify that `reduce` dispatches correctly through each of the typed array code paths — they're special-cased in Clojure's internals. That said, the bug in Q3 is a fair indication that these weren't all carefully verified, so your instinct to question whether we need them all is reasonable.

dgr:
Thanks for the response. Much appreciated. Some follow-up questions.

> 1 & 2. The `interop` map just bundles all the dialect-specific reader conditionals in one place. It would be equally valid as standalone functions but this felt tidier. The purpose of `:int-new` is to construct a boxed `Integer` — you need it because Clojure numeric literals default to `Long`, but `int-array` needs `Integer` objects.

But if you specify the creation of an int array, you get an int array and not a long array, right? That is, you could easily do `(int-array [... ])` and everything is cast appropriately and stored in ints. And you can even do it using `into-array`; see below.

> 3. Yep, that's a bug. Should be `Float/TYPE`. Apologies this got through.  Not sure how.

OK, I'll fix.

> 4. Clojure's persistent vectors are 32-way trees internally. A "node" is one of those tree nodes. `(range 1 100)` gives 99 elements which spans multiple nodes, so we're testing that `reduce` correctly traverses the whole tree rather than just the first 32 elements.

OK, I see.

> 5 & 6. The other typed arrays (`long-array`, `float-array`, `double-array`) accept Clojure's native literal types directly. `int-array` is the awkward one because Clojure's integer literals are `Long` by default, so you need the explicit `(int %)` cast to get the right primitive type.

So: long -> int -> boxed int -> int array slot?

This seems to work just fine, so I'm confused:

```
user=> (into-array Integer/TYPE [1 2 3])
#object["[I" 0x52a70627 "[I@52a70627"]
```

The "[I" in there means it's a Java integer array.

> 7. The intent was to verify that `reduce` dispatches correctly through each of the typed array code paths — they're special-cased in Clojure's internals. That said, the bug in Q3 is a fair indication that these weren't all carefully verified, so your instinct to question whether we need them all is reasonable.

OK

PR #84, opened 2025-06-12
Add tests for empty

rafonsecad:


jeaye (approved):
Nice!

NoahTheDuke:
maybe follow up with a test on a record? that throws an exception in clojure.

PR #85, opened 2025-06-12
Add datatype tests for empty

rafonsecad:
Adding `empty` tests on records as @NoahTheDuke suggested

### DIFF
--- a/test/clojure/core_test/empty.cljc
+++ b/test/clojure/core_test/empty.cljc
@@ -24,4 +24,11 @@
            nil r/max-int
            nil r/min-int
            #?@(:cljs [nil (js/Date)]
-               :clj  [nil (new Object)])))))
+               :clj  [nil (new Object)])))
+
+    (testing "datatypes"
+      (defrecord Record [field])
+      (deftype Type [field])
+      (is (= nil (empty (->Type ""))))
+      #?@(:cljs [(is (= nil (empty (->Record ""))))]
+          :clj  [(is (thrown? UnsupportedOperationException (empty (->Record ""))))]))))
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/empty.cljc
```diff
@@ -24,4 +24,11 @@
            nil r/max-int
            nil r/min-int
            #?@(:cljs [nil (js/Date)]
-               :clj  [nil (new Object)])))))
+               :clj  [nil (new Object)])))
+
+    (testing "datatypes"
+      (defrecord Record [field])
+      (deftype Type [field])
```
jeaye, line 31: Each test should try to use as few other `clojure.core` vars as possible, since those vars may not exist on whichever dialect is running test. On top of that, I'm not sure if every major Clojure dialect implements `defrecord` and `deftype`.

I think it's good thinking to add these, but I'm currently thinking this is doing too much for the `empty` test. Anyone have thoughts on that?

@rafonsecad @NoahTheDuke

rafonsecad: yeah, that could be too much for `empty` test. In the other hand, we could add `when-var-exists` for `defrecord` and `deftype` to avoid the test from failing if those datatypes are not yet implemented

jeaye: Good thoughts, folks. Yeah, let's use a `when-var-exists` for each of these vars. Mainly, I want these tests to also serve as an onboarding ramp for new dialects (like jank), so ideally each test has the smallest footprint it can.

rafonsecad: @jeaye ok I just updated the pull request. Let me know if we need another change

rafonsecad: @jeaye  yeah that makes sense. check again please

jeaye (approved):
Looks good to me. @NoahTheDuke?

jeaye:
Thanks, @rafonsecad!

PR #86, opened 2025-06-13
Add tests for ffirst, fnext, last, nfirst, and nnext

rafonsecad:


test/clojure/core_test/ffirst.cljc
```diff
@@ -0,0 +1,31 @@
...
+      (is (= nil (ffirst [])))
+      (is (= nil (ffirst {})))
+      (is (= nil (ffirst #{})))
+      (is (= nil (ffirst nil)))
+      (is (= :a (ffirst {:a :b})))
+      (is (= 0 (ffirst [[0 1] [2 3]])))
+      (is (= 0 (ffirst '([0 1] [2 3]))))
+      (is (= \a (ffirst ["ab" "cd"])))
+      (is (= \a (ffirst ["abcd"])))
+      (is (= \a (ffirst #{"abcd"}))))
```
jeaye, line 19: I think it'd be good to cover a happy case for infinite seqs. For example, two come to mind:

* `(ffirst (repeat (range)))`
* `(ffirst [(range)])`

Same with `nfirst`.

jeaye (changes requested):
Looking good! Just one request for another lazy case.

test/clojure/core_test/ffirst.cljc
rafonsecad: @jeaye for `ffirst` those cases are ok, but for `nfirst` would work with a finite range
```clojure
(nfirst (repeat (range 0 10))) ;; (1 2 3 4 5 6 7 8 9)

(nfirst (repeat (range))) ; is equal to (next (range))
```

jeaye: Yep, do whichever tweaks you need for `nfirst`. :slightly_smiling_face:

rafonsecad:
@jeaye,  I updated the pull request. Let me know if we need another change

jeaye (approved):
Looks great!

PR #87, opened 2025-06-17
Add tests for hash-map, hash-set, and set

rafonsecad:


test/clojure/core_test/set.cljc
```diff
@@ -0,0 +1,23 @@
+(ns clojure.core-test.set
+  (:require clojure.core
+            [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
+
+(when-var-exists clojure.core/set
+  (deftest test-set
```
jeaye, line 7: I think we should have a test which asserts that `set` removes duplicates.

Something like `(set [1 2 3 12 3])`.

NoahTheDuke:
Maybe add nested calls? So `(= {:a {:b {:c 1} :d 2}} (hash-map :a (hash-map :b (hash-map :c 1) :d 2)))`, etc.

I don't know how exactly to test it, but for example, you can't write `#{(rand) (rand)}` but you can write `(hash-set (rand) (rand))`, so maybe a test for that behavior?

jeaye:
> Maybe add nested calls? So `(= {:a {:b {:c 1} :d 2}} (hash-map :a (hash-map :b (hash-map :c 1) :d 2)))`, etc.

Sounds like a good thing to test.

> I don't know how exactly to test it, but for example, you can't write `#{(rand) (rand)}` but you can write `(hash-set (rand) (rand))`, so maybe a test for that behavior?

We don't want to be using `rand` at all in these tests. Also, the lack of support for `#{(foo) (foo)}` is a *reader* feature, not a `clojure.core` feature. We don't need to test for it here.

rafonsecad:
@NoahTheDuke, thanks for the comment, I just added `(is (= {:a {:b {:c 1} :d 2}} (hash-map :a (hash-map :b (hash-map :c 1) :d 2))))`

jeaye (approved):
Great job! Thanks so much for helping with these.

PR #90, opened 2025-07-04
parse functions

rafonsecad:


test/clojure/core_test/parse_boolean.cljc
```diff
@@ -0,0 +1,38 @@
...
+    (testing "common"
+      (are [expected x] (= expected (parse-boolean x))
+           false "false"
+           true  "true"
+           nil   ""
+           nil   "foo"
+           nil   "False"
+           nil   "FALSE"
+           nil   "True"
+           nil   "TRUE"))
```
jeaye, line 17: These parsing functions are quite involved, so I think we can have quite a bit more rigor here.

Would be good to have edge case tests for prefix/suffix (i.e. `ttrue` and `truee`). Also good to have tests for number strings (since `1` is commonly convertible to bool). Also, we should test for how whitespace is handled (on either end, and in the middle).

test/clojure/core_test/parse_uuid.cljc
```diff
@@ -0,0 +1,34 @@
...
+            [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
+
+(when-var-exists clojure.core/parse-uuid
+  (deftest test-parse-uuid
+    (testing "common"
+      (are [expected x] (= expected (parse-uuid x))
+           nil ""
+           nil "0"
+           nil "df0993")
```
jeaye, line 12: What if it's a valid UUID + some extra?

test/clojure/core_test/parse_double.cljc
```diff
@@ -0,0 +1,43 @@
...
+  (:require clojure.core
+            [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
+
+(when-var-exists clojure.core/parse-double
+  (deftest test-parse-double
+    (testing "common"
+      (are [expected x] (= expected (parse-double x))
+           nil ""
+           nil "foo"
```
jeaye, line 11: Clojure also supports scientific notation, like `1.5e10`, which also supports uppercase `E`.

PR #93, opened 2025-08-27
Update tests for ClojureCLR

dmiller:


test/clojure/core_test/eq.cljc
```diff
@@ -118,6 +118,17 @@
               ;; https://clojure.org/guides/equality notes that sometimes 
               ;; collections with ##NaN are eq
               (list ##NaN) (list ##NaN) true))
+     :cljr (testing "jvm"
```
jeaye, line 121: I think we want `cljr` for the string here.

test/clojure/core_test/add_watch.cljc
```diff
@@ -17,7 +17,7 @@
                       (let [do-update (fn [x]
                                         (try
                                           (swap! x inc)
-                                          (catch #?(:cljs :default :clj clojure.lang.ExceptionInfo) e
+                                          (catch #?(:cljs :default :default clojure.lang.ExceptionInfo) e
```
jeaye, line 20: For every other `Exception` type, we introduced `:cljr`, but for these three in this file, we use `:default`. Let's please use `:cljr`. I'm hoping to phase out all `:default` branches of reader conditionals, since interop can really have no cross-dialect default.

jeaye (changes requested):
Thank you, David! This is excellent work. I have just a couple of notes, but otherwise everything looks ready to merge.

test/clojure/core_test/eq.cljc
dmiller: Definitely.

test/clojure/core_test/add_watch.cljc
jeaye: Yes, I can do these while merging. Will do tomorrow.

dmiller:
Glad you made it through the unbalanced parens.  Obviously I should have done a better final pass after all the little edits.

jeaye:
> Glad you made it through the unbalanced parens. Obviously I should have done a better final pass after all the little edits.

No problem! In terms of getting this hooked into CI, we have a pretty simple setup here: https://github.com/jank-lang/clojure-test-suite/blob/main/.github/workflows/ci.yml#L54-L61

What would it take to get a working Clojure CLR setup running in there?

dmiller:
One person did set up a workflow for testing in one of the libraries I maintain.  Thanks to him, we have a working examples.  Check out: https://github.com/clojure/clr.tools.namespace/blob/master/.github/workflows/test.yml

PR #96, opened 2025-09-02
Add tests for `realized?`

daveliepmann:
Adds tests for `realized?`.

The nature of futures made me sleep the main thread a few times. In theory this could be brittle timing-wise but it should be fine in context. A cross-platform `sleep` also seemed useful.


jeaye (approved):
Excellent! Thanks for the iteration, Dave.

PR #102, opened 2025-09-05
test `atom`

### LINKED ISSUE #155: `clojure.core/atom`

daveliepmann:
Tests `atom`.

Fixes #155

### DIFF
--- a/test/clojure/core_test/atom.cljc
+++ b/test/clojure/core_test/atom.cljc
@@ -70,17 +70,24 @@
           (is (= 3 (deref atm)))))
 
       (testing "always-falsey validator can't initialize atom"
-        ;; CLJS currently fails due to a bug https://clojure.atlassian.net/browse/CLJS-3447
-        (is (thrown? #?(:cljs :default, :default Exception)
-                     (atom {} :validator (constantly nil))))
-        (is (thrown? #?(:cljs :default, :default Exception)
-                     (atom {} :validator (constantly false))))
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/atom.cljc
```diff
@@ -0,0 +1,142 @@
...
+        \newline
+        nil
+        true
+        false
+        ##Inf
+        :kw
+        :ns/kw
+        '(one two three)
+        [1 2 3]
+        {:k "value"}
```
E-A-Griffin, line 136: I would add a regex literal here for completeness

daveliepmann:
Digging into some odd CLJS validator behavior

test/clojure/core_test/atom.cljc
jeaye: I think we can pass on this, due to the oddities of regex equality and the unrelatedness to `atom` in particular.

daveliepmann:
CLJS bug discovered: https://clojure.atlassian.net/browse/CLJS-3447

daveliepmann:
CLJS currently fails x4, which is correct behavior because of the [CLJS-3447](https://clojure.atlassian.net/browse/CLJS-3447) bug this uncovered. How do we handle failures we want to happen?

jeaye:
Side note, I've updated the description to "Fixes #155", since that will lead to the ticket automatically closing on merge. GH doesn't do that for "Solves". I think it's just "Fixes" and "Closes" that it supports.

PR #103, opened 2025-09-05
Test `constantly`

daveliepmann:
Tests `constantly`

### DIFF
--- a/test/clojure/core_test/constantly.cljc
+++ b/test/clojure/core_test/constantly.cljc
@@ -23,7 +23,7 @@
         '(one two three)
         [4 5 6]
         {:7 "8"}
-        (zipmap (take 1000 (range))
+        (zipmap (take 100 (range))
                 (cycle ['foo 'bar 'baz 'qux]))
         #{:a :b "c"})
 
@@ -40,5 +40,5 @@
         (is (= :foo (the-fn {:k :v})))
         (is (= :foo (the-fn even?)))
         (is (= :foo (the-fn (some-fn even? odd?))))
-        (is (= :foo (apply the-fn (take 10000 (cycle ['foo "bar" :baz])))))
+        (is (= :foo (apply the-fn (take 1000 (cycle ['foo "bar" :baz])))))
         (is (apply = (map (constantly (rand-int 100)) [:a :b :c :d :e :f])))))))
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/constantly.cljc
```diff
@@ -0,0 +1,44 @@
...
+  (deftest test-constantly
+    (testing "`constantly`"
+      (is (fn? (constantly nil)))
+
+      (are [v] (= v ((constantly v)))
+        `sym
+        "abc"
+        100000
+        1.0
+        #?(:clj 2/3) ; ratio not supported in cljs
```
E-A-Griffin, line 16: Should this be `:default`? I assume most Clojure variants support ratios and this is basically just a cljs issue

test/clojure/core_test/constantly.cljc
```diff
@@ -0,0 +1,44 @@
...
+        \return
+        nil
+        true
+        false
+        :keyword
+        :namespace/keyword
+        '(one two three)
+        [4 5 6]
+        {:7 "8"}
+        (zipmap (take 1000 (range))
```
E-A-Griffin, line 26: Get what you're going for but 1000 may be overkill, I think 100 would probably be plenty, I know this doesn't make a noticeable difference in performance but if everyone adds a 1k range to each test then it probably will accumulate to a noticeable increase in test execution time

test/clojure/core_test/constantly.cljc
jeaye: Yep, I think CLJS is unique in not supporting ratios.

E-A-Griffin: I think we'd need an explicit `:cljs` for `:default` to work properly but we could just put something else here for `:cljs`

E-A-Griffin: I would do a `#js` object here for `:cljs` instead of the ratio

E-A-Griffin: could get cheeky and do
```clj
:cljs #js {"cljs does not have" "ratios"}
```

jeaye (changes requested):
Thanks, Dave! After Emma's comments are addressed, this looks great to me.

daveliepmann:
I think we're good to go, thanks for the close read :)

PR #107, opened 2025-09-05
Additional edits for ClojureCLR tests

dmiller:


dmiller:
The failing test in shuffle does not happen on my system.

You'll want to take a close at the changes.  I had to do some non-trivial rearrangement in the string library tests.

Also, for with-out-str, I had to introduce a helper function from clojure.test-helpers.  I could not figure out how to :require or :use that library.  (Both ClojureJVM and ClojureCLR use it in their test suites.)

deps-clr.edn
```diff
@@ -0,0 +1,11 @@
+{:deps {}
+ 
+   :aliases
```
E-A-Griffin, line 3: weird spacing and indentation

test/clojure/core_test/byte.cljc
```diff
@@ -13,16 +13,17 @@
...
       0    0
       127  127
       1    1N
       0    0N
-      -1   -1N
+      #?@(:cljr [] :default [-1   -1N])
       1    1.0M
       0    0.0M
-      -1   -1.0M
+      #?@(:cljr [] :default [-1   -1.0M])
```
E-A-Griffin, line 26: spacing is weird

test/clojure/core_test/byte.cljc
```diff
@@ -34,14 +35,14 @@
...
            1    1.9
            1    3/2
-           -1   -3/2
+           #?@(:cljr [] :default  [-1   -3/2])
            0    1/10
-           0    -1/10
+           #?@(:cljr [] :default [0    -1/10])
            1    1.1M
-           -1   -1.1M]))
+           #?@(:cljr [] :default [-1   -1.1M])]))
```
E-A-Griffin, line 45: spacing is weird for each of these

test/clojure/core_test/with_out_str.cljc
```diff
@@ -2,10 +2,24 @@
...
 
+;; This is part of clojure.test-helpers, but I couldn't figure out how to :require or :use that library.
+;; Copied here for now
+
+#?(:cljr
+   (let [nl Environment/NewLine]                                       ;;; (System/getProperty "line.separator")] 
+     (defn platform-newlines [s] (.Replace ^String s "\n" nl)))        ;;; .replace, add type hint
+   :clj
+   (let [nl (System/getProperty "line.separator")] 
+    (defn platform-newlines [s] (.replace s "\n" nl)))
```
E-A-Griffin, line 13: style nitpick but I think it'd be cleaner to have the var defined with a `defn` at the top level for all implementations and then change the body based on platform, i.e.
```clj
(defn platform-newlines [s]
  #?(:clj
     (let [nl (System/getProperty "line.separator")]
       (.replace s "\n" nl))
     :cljr
     (let [nl Environment/NewLine]           ;;; (System/getProperty "line.separator")] 
       (.Replace ^String s "\n" nl)))        ;;; .replace, add type hint
     :default
     s)

File: deps-clr.edn [hunk 5]
```diff
@@ -0,0 +1,11 @@
...
+   {:test
+     {:extra-paths ["test"]
+	  :extra-deps {io.github.dmiller/test-runner {:git/sha "c055ea13d19c6a9b9632aa2370fcc2215c8043c3"}}
+	  ;; :main-opts {"-m" "cognitect.test-runner" "-d" "test"}
+	  :exec-fn cognitect.test-runner.api/test
+	  :exec-args {:dirs ["test"]
+	              :patterns [".*test.*"]}}}
```
E-A-Griffin, line 10: I don't think this will work, most namespaces don't have `test` in the name

deps-clr.edn
dmiller: fix as you wish.
a copy of a copy of a copy from somewhere someone else used the test-runner

test/clojure/core_test/byte.cljc
```diff
@@ -13,16 +13,17 @@
...
+	;; In ClojureCLR, Byte is unsigned, so we have to wipe all tests of negative values
     (are [expected x] (= expected (byte x))
-      -128 -128
+      #?@(:cljr [] :default [-128 -128])
       0    0
       127  127
       1    1N
       0    0N
-      -1   -1N
+      #?@(:cljr [] :default [-1   -1N])
```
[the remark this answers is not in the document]
dmiller: I had just inserted the original code in the :default case.
Format as you see fit, for sure.

test/clojure/core_test/with_out_str.cljc
dmiller: agreed

deps-clr.edn
dmiller: you are probably right, though i think all tests were running for me. I think you can just leave out the :patterns.

E-A-Griffin: I'll try it again but it didn't work for me when I tried this last night

E-A-Griffin:
@dmiller this is what I got
```
Ran 172 tests containing 3949 assertions.
40 failures, 24 errors.
```

deps-clr.edn
E-A-Griffin: this works, thanks!

E-A-Griffin (approved):
There's still some failing CLR tests (on my end, not sure if @dmiller has any failures)  but I think we can clean those up later, I'm happy to take a further look in a follow up but this is blocking multiple PRs so I think better to get it through and make incremental progress

dmiller:
That sounds like a plan.
I'm on call to help resolve issues.
At present I have only seen intermittent failure in tests in add-watch and
tap.
Given that they are intermittent, I'm wondering about some timing problems
in those tests.
But otherwise, I have only passes.

E-A-Griffin:
@dmiller if the tests are passing on your end then it's probably an issue on my local

E-A-Griffin:
@jeaye okay, this time for real I think it's ready

PR #751, opened 2025-09-06
test `var?`

daveliepmann:
addresses #561

### DIFF
--- a/test/clojure/core_test/var_qmark.cljc
+++ b/test/clojure/core_test/var_qmark.cljc
@@ -10,10 +10,10 @@
 (when-var-exists clojure.core/defmulti
   ;; This can't be inside `deftest` like `defn` because `defmulti` only returns
   ;; the var on the first invocation.
-  (var? (defmulti bar first)))
+  (defmulti bar first))
 
 (when-var-exists clojure.core/defprotocol
-  (var? (defprotocol MyProtocol)))
+  (defprotocol MyProtocol))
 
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/var_qmark.cljc
```diff
@@ -0,0 +1,71 @@
...
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
+
+(def foo :foo)
+
+(def ^:dynamic i-am-dynamic 3.14)
+
+(when-var-exists clojure.core/defmulti
+  ;; This can't be inside `deftest` like `defn` because `defmulti` only returns
+  ;; the var on the first invocation.
+  (var? (defmulti bar first)))
```
E-A-Griffin, line 13: should this be an `assert` then if it's not able to be an `is`?

test/clojure/core_test/var_qmark.cljc
```diff
@@ -0,0 +1,71 @@
+(ns clojure.core-test.var-qmark
+  (:require clojure.core
```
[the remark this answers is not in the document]
daveliepmann: It's auto-generated by `bb new-test`

test/clojure/core_test/var_qmark.cljc
daveliepmann: Whoops, those `var?` are cruft — they are tested with `is` but the definitions have to happen prior.

PR #752, opened 2025-09-06
Fix whitespace in template

E-A-Griffin:
Removes superfluous whitespace in template


E-A-Griffin:
Just moving this to the other PR since this is so small and similar https://github.com/jank-lang/clojure-test-suite/pull/753

PR #760, opened 2025-09-21
Fix bad `:default` min value

djblue:
This test was failing for me in jank and I think this test value is incorrect.

```clojure
% clj
Clojure 1.12.2
user=> Long/MIN_VALUE
-9223372036854775808
user=> -0x8000000000000000
-9223372036854775808
```

### DIFF
--- a/test/clojure/core_test/number_range.cljc
+++ b/test/clojure/core_test/number_range.cljc
@@ -10,7 +10,7 @@
 (def ^:const min-int #?(:clj Long/MIN_VALUE
                         :cljr Int64/MinValue
                         :cljs js/Number.MIN_SAFE_INTEGER
-                        :default 0x8000000000000000))
+                        :default -0x8000000000000000))
 
 (def ^:const all-ones-int #?(:cljs 0xFFFFFFFF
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/number_range.cljc
```diff
@@ -10,7 +10,7 @@
 (def ^:const min-int #?(:clj Long/MIN_VALUE
                         :cljr Int64/MinValue
                         :cljs js/Number.MIN_SAFE_INTEGER
-                        :default 0x8000000000000000))
+                        :default -0x8000000000000000))
```
jeaye, line 13: On which host was this failing? bb? Every other host we support is covered above.

jeaye: I think we should remove this constant. It's all sorts of weird. We should just add a `:jank` branch which does `(cpp/value "std::numeric_limits<jtl::i64>::min()")`.

jeaye:
@shantanu-sardesai Will you please also get this merged, in a separate PR? It's related to your latest PR here.

PR #762, opened 2025-09-29
test `mapcat` (#354)

gavlooth:

PR #763, opened 2025-10-15
Test mapcat

E-A-Griffin:
Addresses comments in https://github.com/jank-lang/clojure-test-suite/pull/762#issue-3463516980 for @gavlooth 's work; fixes/closes https://github.com/jank-lang/clojure-test-suite/issues/354#issue-3388609679

### DIFF
--- a/test/clojure/core_test/mapcat.cljc
+++ b/test/clojure/core_test/mapcat.cljc
@@ -56,7 +56,4 @@
                    (mapcat 42 [1 2]))))
     (testing "non-concatable return value"
       (is (thrown? #?(:cljs :default :default Exception)
-                   ;; we need to realize the lazy data structure
-                   ;; in order for `cljs` to throw on the result
-                   ;; hence `(into [])`
-                   (into [] (mapcat identity (range 2))))))))
+                   (doall (mapcat identity (range 2))))))))


E-A-Griffin:
@jeaye on the local cljs repls I have I get an error for `(mapcat identity (range 2))` any idea why the CI suite gets a different result?

E-A-Griffin:
```
clojure.core-test.mapcat> (emma-test)

FAIL in (emma-test) (:1:37)
x
expected: (thrown? :default (mapcat identity (range 2)))
  actual: nil
```
Okay reproduced it with cider, I'll look into this tonight

E-A-Griffin:
```
(deftest emma-test2 (testing "y" (is (nil? (mapcat identity (range 2))))))
;; => #'clojure.core-test.mapcat/emma-test2
cljs.user> (in-ns 'clojure.core-test.mapcat)
;; => nil
clojure.core-test.mapcat> (emma-test2)

FAIL in (emma-test2) (:1:38)
y
expected: (nil? (mapcat identity (range 2)))

ERROR in (emma-test2) (:1:38)
y
expected: (nil? (mapcat identity (range 2)))
  actual: #object[Error Error: 0 is not ISeqable]
;; => nil
```
So it looks like there's a certain class of error that isn't being caught for some reason?

project.clj
```diff
@@ -3,5 +3,6 @@
   :url "https://github.com/jank-lang/clojure-test-suite"
   :license {:name "MPL 2.0"
             :url "https://www.mozilla.org/en-US/MPL/2.0/"}
-  :dependencies [[org.clojure/clojure "1.12.0"]]
+  :dependencies [[org.clojure/clojure "1.12.0"]
+                 [org.clojure/clojurescript "1.12.42"]]
```
E-A-Griffin: cider doesn't work on this project without this, @jeaye if you want me to do this in a separate PR that's fine but I think we should have it

test/clojure/core_test/mapcat.cljc
```diff
@@ -56,4 +56,7 @@
                    (mapcat 42 [1 2]))))
     (testing "non-concatable return value"
       (is (thrown? #?(:cljs :default :default Exception)
-                   (mapcat identity (range 2)))))))
+                   ;; we need to realize the lazy data structure
+                   ;; in order for `cljs` to throw on the result
+                   ;; hence `(into [])`
+                   (into [] (mapcat identity (range 2))))))))
```
jeaye: This is what `doall` is for.

E-A-Griffin: Do you prefer `doall` to realizing a vector, they both work here (I don't have a preference)?

jeaye: Yes, please use `doall`. We don't want a vector, we just want to force the lazy seq to be realized. That is the exact purpose of `doall`. That also allows us to remove the comment, since `doall` being single purpose makes its usage self-documenting.

jeaye:
Nice work, Emma and Christos!

PR #764, opened 2025-10-16
pop tests

rafonsecad:


E-A-Griffin:
Can you add an assertion for `pop`'s behavior with `clojure.lang.PersistentQueue`? Most Clojure dialects support this, `cljs` is the only one I know of that doesn't

E-A-Griffin:
`clojure.lang.MapEntry` also implements `pop` but that feels more obscure so I'll defer to @jeaye and you on whether we need a test for that; I don't have a strong opinion.

jeaye:
> Can you add an assertion for `pop`'s behavior with `clojure.lang.PersistentQueue`? Most Clojure dialects support this, `cljs` is the only one I know of that doesn't

Does Clojure CLR have this? I think we should focus more on `clojure.core` and not the internals of Clojure JVM.

> clojure.lang.MapEntry also implements pop but that feels more obscure so I'll defer to @jeaye and you on whether we need a test for that; I don't have a strong opinion.

No, we should focus on `pop`, not all of the details of Clojure JVM's object model. Some Clojure dialects don't have a dedicated map entry type, since a vector of two suffices.

E-A-Griffin:
> > Can you add an assertion for `pop`'s behavior with `clojure.lang.PersistentQueue`? Most Clojure dialects support this, `cljs` is the only one I know of that doesn't
> 
> Does Clojure CLR have this? I think we should focus more on `clojure.core` and not the internals of Clojure JVM.
> 
> > clojure.lang.MapEntry also implements pop but that feels more obscure so I'll defer to @jeaye and you on whether we need a test for that; I don't have a strong opinion.
> 
> No, we should focus on `pop`, not all of the details of Clojure JVM's object model. Some Clojure dialects don't have a dedicated map entry type, since a vector of two suffices.

Clojure CLR does have this, I would assume any public data structure in `clojure.lang.*` would be intended behavior rather than JVM implementation details

E-A-Griffin:
Looking at other Clojure platforms it looks like
**Has `clojure.lang.PersistentQueue`**
- Clojure
- bb
- ClojureCLR

**Doesn't have `clojure.lang.PersistentQueue`**
- ClojureScript
- nbb
- Clojerl
- ClojureDart (didn't test this one but saw some tests commented out from skimming)

jeaye:
> Looking at other Clojure platforms it looks like **Has `clojure.lang.PersistentQueue`**
> 
>     * Clojure
> 
>     * bb
> 
>     * ClojureCLR
> 
> 
> **Doesn't have `clojure.lang.PersistentQueue`**
> 
>     * ClojureScript
> 
>     * nbb
> 
>     * Clojerl
> 
>     * ClojureDart (didn't test this one but saw some tests commented out from skimming)

jank will be in the latter list as well. To me, `clojure.lang.*` is an implementation detail of Clojure JVM. bb gets it for free, so really only Clojure CLR did the work to implement this.

rafonsecad:
I replaced `:clj` and `:cljr` with `:default` as @E-A-Griffin suggested. Let me know if we need to change something else

E-A-Griffin (approved):

Issue #201, opened 2025-09-05
`clojure.core/conj`


jeaye:

rafonsecad:
I'd like to create the tests for this one

PR #768, opened 2025-10-24
test conj

rafonsecad:
closes https://github.com/jank-lang/clojure-test-suite/issues/201

### DIFF
--- a/test/clojure/core_test/conj.cljc
+++ b/test/clojure/core_test/conj.cljc
@@ -5,6 +5,7 @@
 (when-var-exists conj
   (deftest test-conj
     (testing "common"
+      (is (= [] (conj)))
       (is (= nil (conj nil)))
       (is (= '(nil) (conj nil nil)))
       (is (= {} (conj {})))


test/clojure/core_test/conj.cljc
```diff
@@ -0,0 +1,33 @@
+(ns clojure.core-test.conj
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists]]))
+
+(when-var-exists conj
+  (deftest test-conj
```
jeaye: I think we can have a test for adding to the front of an infinite seq. For example `(conj (range) -1))` is valid.

test/clojure/core_test/conj.cljc
```diff
@@ -0,0 +1,33 @@
...
+      (is (= #{} (conj #{})))
+      (is (= '() (conj '())))
+      (is (= '(3) (conj nil 3)))
+      (is (= '([1 2]) (conj nil [1 2])))
+      (is (= [1 2 3 4] (conj [1 2 3] 4)))
+      (is (= '(4 1 2 3) (conj '(1 2 3) 4)))
+      (is (= '(\f \e \d \a \b \c) (conj '(\a \b \c) \d \e \f)))
+      (is (= [[1 2] [3 4] [5 6]] (conj [[1 2] [3 4]] [5 6])))
+      (is (= {:a 0 :b 1} (conj {:a 0} [:b 1])))
+      (is (= {:a 0 :b 1} (conj {:a 0} {:b 1})))
```
jeaye: Existing map keys can also be updated. `(conj {:a 0} {:a 1})`, for example.

E-A-Griffin (commented):
The only thing I'd suggest adding is a test for the 0-arity call to `conj`

rafonsecad:
I just added the 0-arity test, thanks @E-A-Griffin

E-A-Griffin (approved):
Looks good to me if @jeaye also thinks it's good to go, thanks for your work on this!

Issue #1, opened 2024-10-03
Test all the things :tada: 


jeaye:
- [ ] `*`
 - [ ] `*'`
 - [ ] `*1`
 - [ ] `*2`
 - [ ] `*3`
 - [ ] `*agent*`
 - [ ] `*allow-unresolved-vars*`
 - [ ] `*assert*`
 - [ ] `*clojure-version*`
 - [ ] `*command-line-args*`
 - [ ] `*compile-files*`
 - [ ] `*compile-path*`
 - [ ] `*compiler-options*`
 - [ ] `*data-readers*`
 - [ ] `*default-data-reader-fn*`
 - [ ] `*e`
 - [ ] `*err*`
 - [ ] `*file*`
 - [ ] `*flush-on-newline*`
 - [ ] `*fn-loader*`
 - [ ] `*in*`
 - [ ] `*math-context*`
 - [ ] `*ns*`
 - [ ] `*out*`
 - [ ] `*print-dup*`
 - [ ] `*print-length*`
 - [ ] `*print-level*`
 - [ ] `*print-meta*`
 - [ ] `*print-namespace-maps*`
 - [ ] `*print-readably*`
 - [ ] `*read-eval*`
 - [ ] `*reader-resolver*`
 - [ ] `*source-path*`
 - [ ] `*suppress-read*`
 - [ ] `*unchecked-math*`
 - [ ] `*use-context-classloader*`
 - [ ] `*verbose-defrecords*`
 - [ ] `*warn-on-reflection*`
 - [ ] `+`
 - [ ] `+'`
 - [ ] `-`
 - [ ] `-'`
 - [ ] `->`
 - [ ] `->>`
 - [ ] `->ArrayChunk`
 - [ ] `->Eduction`
 - [ ] `->Vec`
 - [ ] `->VecNode`
 - [ ] `->VecSeq`
 - [ ] `-cache-protocol-fn`
 - [ ] `-reset-methods`
 - [ ] `.`
 - [ ] `..`
 - [ ] `/`
 - [ ] `<`
 - [ ] `<=`
 - [ ] `=`
 - [ ] `==`
 - [ ] `>`
 - [ ] `>=`
 - [x] `abs`
 - [ ] `accessor`
 - [ ] `aclone`
 - [ ] `add-classpath`
 - [ ] `add-tap`
 - [ ] `add-watch`
 - [ ] `agent`
 - [ ] `agent-error`
 - [ ] `agent-errors`
[...]

PR #770, opened 2025-10-28
All remaining clojure.core/v* and w* public functions

VaclavSynacek:
I understand this is potenciality controversial. But do we really need the tests written by humans? I tweaked a bit the instructions from #1 and let AI run on it. I reviewed some parts and to my eyes it is 90% of what a human would do (maybe some parts a bit redundant and over tested), much better than no tests, maybe some exceptional humans can add a few extra tricky tests (not me). It passes on clojure as well as clojurescript and mostly avoids reader conditionals if possible.

I understand some projects have legitimate reasons to not allow any AI code in. In this case IMHO risk is minimal and we can get all tests written for almost free.

So here are all remaining clojure.core functions starting with 'v' and 'w'. I can bring more if this is ok.

### DIFF
--- a/test/clojure/core_test/when_first.cljc
+++ b/test/clojure/core_test/when_first.cljc
@@ -1,0 +1,98 @@
+(ns clojure.core-test.when-first
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists]]))
+
+(when-var-exists when-first
+  (deftest test-when-first
+    (testing "binds first element and executes"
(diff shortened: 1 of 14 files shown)


VaclavSynacek:
> [#1 (comment)](https://github.com/jank-lang/clojure-test-suite/issues/1#issuecomment-3458218688)

Ok, fair enough, your analysis is probably correct - it seemed like sight overtesting to me too (which I misjudged as reasonable tradeoff for having full test suite sooner). Maybe this can be mitigated partially by refining the prompts to generate less code and double check for overlapping tests - but it can hardly get to the point of handwritten code you wish to maintain. Sorry I could not have helped.

jeaye:
No problem. Thank you for your efforts!

PR #791, opened 2025-11-07
Test special-symbol?

### LINKED ISSUE #498: `clojure.core/special-symbol?`

super7ramp:
This closes #498.


test/clojure/core_test/special_symbol_qmark.cljc
```diff
@@ -0,0 +1,51 @@
...
+                 quote
+                 recur
+                 set!
+                 throw
+                 try
+                 var))
+
+    (testing "more special symbols"
+      #?(:cljs    "cljs doesn't have these"
+         :default (are [arg] (special-symbol? 'arg)
```
jeaye: ~~Isn't this just doing `(special-symbol? (quote arg))` for all of these?~~

EDIT: I think we can remove this section. Not every dialect will have these specials and we don't know which ones will. But it doesn't give us any increased confidence in the fn working, since the above tests are sufficient.

PR #792, opened 2025-11-07
Test vec

### LINKED ISSUE #563: `clojure.core/vec`

super7ramp:
This closes #563.

### DIFF
--- a/test/clojure/core_test/vec.cljc
+++ b/test/clojure/core_test/vec.cljc
@@ -1,0 +1,37 @@
+(ns clojure.core-test.vec
+  (:require [clojure.test :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists]]))
+
+(when-var-exists vec
+  (deftest test-vec
+
+    (testing "vec coll"
+      (are [expected coll] (= expected (vec coll))
+                           [] nil
+                           [] '()
+                           [] []
+                           [] #{}
+                           [] {}
+                           [] ""
+                           [nil nil nil] '(nil nil nil)
+                           [1 2 3] '(1 2 3)
+                           [1 2 3] [1 2 3]
+                           [1 2 3] (sorted-set 1 2 3)
+                           [1 2 3] (range 1 4)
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/vec.cljc
```diff
@@ -0,0 +1,37 @@
...
+                           [1 2 3] (range 1 4)
+                           [[:a 1] [:b 2]] {:a 1 :b 2}
+                           [\a \b \c] "abc"))
+
+    #?(:cljr    "cljr does not alias array"
+       :default (testing "array aliasing"
+                  (let [arr (to-array [1 2 3]), v (vec arr)]
+                    (is (= [1 2 3] v))
+                    (aset arr 0 -1)
+                    (is (= [-1 2 3] v)))))
```
E-A-Griffin: Small nitpick, even though it's not required, it might be clearer that this mutation and the assertion are related if they were in a `do` block or logically grouped in some other way like `doto`). Not a big thing just weird to see mutating code in Clojure without a `!` suffix, but I won't block on this

super7ramp: I agree this mutable code is surprising but I don't see how to test this case differently. I wonder why `aset` is `aset` and not `aset!` 😅

clj-kondo warns of "redundant do" when adding a `do` under `let`.

I don't see how to apply `doto` here 🤔

E-A-Griffin (approved):

test/clojure/core_test/vec.cljc
```diff
@@ -0,0 +1,37 @@
...
+                           [nil nil nil] '(nil nil nil)
+                           [1 2 3] '(1 2 3)
+                           [1 2 3] [1 2 3]
+                           [1 2 3] (sorted-set 1 2 3)
+                           [1 2 3] (range 1 4)
+                           [[:a 1] [:b 2]] {:a 1 :b 2}
+                           [\a \b \c] "abc"))
+
+    #?(:cljr    "cljr does not alias array"
+       :default (testing "array aliasing"
```
jeaye: Very neat test case.

PR #794, opened 2025-11-10
Test persistent!

### LINKED ISSUE #394: `clojure.core/persistent!`

super7ramp:
This closes #394.


E-A-Griffin (approved):

test/clojure/core_test/persistent_bang.cljc
```diff
@@ -0,0 +1,34 @@
...
+                           [] (transient [])
+                           [nil] (transient [nil])
+                           [1 2 3] (transient [1 2 3])))
+
+    (testing "set"
+      (are [expected coll] (= expected (persistent! coll))
+                           #{} (transient #{})
+                           #{nil} (transient #{nil})
+                           #{:a :b :c} (transient #{:a :b :c})))
+
```
jeaye: I think we'll want a test for calling `persistent!` multiple times on a transient, since that will throw. We don't need to do it for every possible shape, but we do want to capture that behavior, since it's documented.

PR #795, opened 2025-11-10
Prep `double?` for `jank`

shantanu-sardesai:
More context can be found [here](https://github.com/jank-lang/jank/pull/583#issuecomment-3503038668).


jeaye:
Yep, thanks!

PR #797, opened 2025-11-11
Test conj!

### LINKED ISSUE #202: `conj!`

super7ramp:
This closes #202.

### DIFF
--- a/test/clojure/core_test/conj_bang.cljc
+++ b/test/clojure/core_test/conj_bang.cljc
@@ -1,0 +1,77 @@
+(ns clojure.core-test.conj-bang
+  (:require [clojure.test :refer [are deftest is testing]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists]]))
+
+(when-var-exists conj!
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/conj_bang.cljc
```diff
@@ -0,0 +1,77 @@
...
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists]]))
+
+(when-var-exists conj!
+  (deftest test-conj!
+
+    (testing "conj!"
+      (is (= [] (persistent! (conj!)))))
+
+    (testing "conj! coll"
+      (testing "when coll is transient"
```
E-A-Griffin: Can we add a test for the count of the transient vector before coercing it to a persistent vector?

super7ramp: Why?

E-A-Griffin: Shouldn't we ensure that the transient has the correct size/shape during intermediate calls to `conj!` (or any other `x!` applied to a transient)?

PR #798, opened 2025-11-13
Cleanup: Fix `core-test` Deps Order, Remove Unused, etc.

E-A-Griffin:
This is a large PR full of a ton of small refactorings, mostly for consistency, some just general cleanup.

Given how large this already is, I'll handle `string-test` in a separate PR

Most changes fit into one of the following, and if they don't I've left a comment to make this easier to review:

1. Reordering `:refer` listing to be alphabetical
2. Cleaning up whitespace inconsistencies
3. Reordering assertions to be visually grouped in similar ways
4. Removing redundant cases that should be covered by `:default` or be treated the same between platforms (most often this is collapsing `:clj` and `:cljr` into `:default`)
5. Removing unused requires/refers
6. Adding `clj-kondo` ignore comments so the files are easier to read for folks using `clj-kondo` and so that actual warnings appear differently and don't get lost
7. Removing unnecessary namespace qualification (e.g. `clojure.core/x` instead of `x`)
8. Changing unused bindings from `x` to `_x`

### DIFF
--- a/test/clojure/core_test/add_watch.cljc
+++ b/test/clojure/core_test/add_watch.cljc
@@ -73,7 +73,7 @@
 
     #?@(:cljs []
         :default
-         (def testvar-a 0)
+        [(def testvar-a 0)
          (def testvar-b 10)
 
          (testing "watch var"
@@ -139,7 +139,7 @@
                (is (contains? (set (keyed :e @state))
                               {:key :e :ref #'testvar-a :old 4 :new 5 :tester :err}))
                (is (contains? (set (keyed :e @state))
-                              {:key :e :ref #'testvar-b :old 14 :new 15 :tester :err}))))))
+                              {:key :e :ref #'testvar-b :old 14 :new 15 :tester :err})))))])
 
     #?(:cljs nil
        :default


test/clojure/core_test/bigint.cljc
```diff
@@ -1,17 +1,17 @@
...
+(when-var-exists bigint
   (deftest test-bigint
     (are [expected x] (= expected (bigint x))
       1N  1
       0N  0
       -1N -1
       #?@(:cljs []
-          :clj [179769313486231570000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000N r/max-double]
-		  :clj [179769313486231570000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000N r/max-double])
+          :cljr [179769313486231570814527423731704356798070567525844996598917476803157260780028538760589558632766878171540458953514382464234321326889464182768467546703537516986049910576551282076245490090389328944075868508455133942304583236903222948165808559332123348274797826204144723168738177180919299881250404026184124858368N r/max-double]
```
E-A-Griffin: Added `cljr` specific largest double

test/clojure/core_test/bound_fn.cljc
```diff
@@ -1,51 +1,51 @@
 (ns clojure.core-test.bound-fn
-  (:require [clojure.test :as t]
-            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer)  [when-var-exists]]))
+  (:require [clojure.test :as t :refer [deftest is testing]]
```
E-A-Griffin: converted to `refer` for consistency with the rest of the project

test/clojure/core_test/bound_fn.cljc
```diff
@@ -1,51 +1,51 @@
...
+         (is (= ((f)) :inside-f) "bound-fn as result preserves initial bindings"))))
 
-   (t/testing "Threaded/future cases"
-     (let [f (bound-fn [] *x*)]
-       (let [fut (future (f))]
-         (binding [*x* :here]
-           (t/is (= @fut :unset) "bound-fn stays bound even in other thread"))))
+   (testing "Threaded/future cases"
+     (let [f (bound-fn [] *x*)
+           fut (future (f))]
```
E-A-Griffin: removed redundant `let`

test/clojure/core_test/format.cljc
```diff
@@ -1,8 +1,9 @@
+#_{:clj-kondo/ignore [:unused-referred-var]}
```
jeaye: This isn't something we should have to put in every file which isn't using a referred var. If we want kondo to be clean, we should disable these diagnostics at a project level. Same goes for things like inline defs and unused values.

PR #808, opened 2025-11-16
Add tests for `when-let`

alysbrooks:
Tests are based on the tests for `when-first`.

### DIFF
--- a/test/clojure/core_test/when_let.cljc
+++ b/test/clojure/core_test/when_let.cljc
@@ -1,5 +1,5 @@
 (ns clojure.core-test.when-let
-  (:require [clojure.test :as t :refer [deftest testing is are]]
+  (:require [clojure.test :as t :refer [are deftest is testing]]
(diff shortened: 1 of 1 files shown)


jeaye:
CLJS does weird stuff with macros. We could just disable that test for CLJS using a reader conditional.

E-A-Griffin (approved):
Thanks @alysbrooks !

jeaye:
> I created an issue for Babashka's behavior, which doesn't seem correct: [babashka/babashka#1894](https://github.com/babashka/babashka/issues/1894).

Great find, Alys. If we're going to skip bb for now, can we please link to the bb issue there to explain why? It would also be helpful to have a comment for the CLJS condition, to mention that CLJS macroexpansion is non-standard so it's being skipped.

alysbrooks:
I switched computers and my Git settings are a little different, which was causing CLA automation to think I was two different people. I think it's fixed now so it's all attributed to my GitHub account.

PR #809, opened 2025-11-16
Add missing symbol to word translation for `-`

shantanu-sardesai:


jeaye (changes requested):

jeaye:
The wording could be much simpler than in this PR, while still following the format we have, so I achieved it here: https://github.com/jank-lang/clojure-test-suite/commit/a28377b800a76b06b8aa93930b1b8a5e1fd2fa83

Thanks for pointing this out!

PR #810, opened 2025-11-16
Test uuid?

### LINKED ISSUE #556: `clojure.core/uuid?`

super7ramp:
This closes #556.

### DIFF
--- a/test/clojure/core_test/uuid_qmark.cljc
+++ b/test/clojure/core_test/uuid_qmark.cljc
@@ -1,0 +1,36 @@
+(ns clojure.core-test.uuid-qmark
+  (:require [clojure.test :refer [are deftest testing]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists]]))
+
+(when-var-exists uuid?
+  (deftest test-uuid?
+
+    (testing "uuids"
+      (are [x] (uuid? x)
+               #uuid "00000000-0000-0000-0000-000000000000"
+               #uuid "f81d4fae-7dec-11d0-a765-00a0c91e6bf6"
+               (parse-uuid "f81d4fae-7dec-11d0-a765-00a0c91e6bf6")
+               (random-uuid)))
+
+    (testing "not uuids"
+      (are [x] (not (uuid? x))
+               nil
+               true
+               false
+               0
+               1
+               -1
+               0.0
+               1.0
+               -1.0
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/uuid_qmark.cljc
```diff
@@ -0,0 +1,39 @@
...
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists]]))
+
+(when-var-exists uuid?
+  (deftest test-uuid?
+
+    (testing "uuids"
+      (are [x] (uuid? x)
+               #uuid "f81d4fae-7dec-11d0-a765-00a0c91e6bf6"
+               (parse-uuid "f81d4fae-7dec-11d0-a765-00a0c91e6bf6")
+               (random-uuid)))
```
E-A-Griffin: Might be a good idea to test the "nil" uuid (i.e. `00000000-0000-0000-0000-000000000000`) occasionally runtimes treat this UUID as a special case, which would be good to catch

super7ramp: Thanks, added case for nil uuid.

NoahTheDuke:
Does using `#uuid` hurt portability? or do we expect all conforming dialects to implement `#uuid`?

jeaye:
> Does using `#uuid` hurt portability? or do we expect all conforming dialects to implement `#uuid`?

I'm expecting all dialects to support it until we have some dialects which don't and then we need to make some hard calls. :joy:

test/clojure/core_test/uuid_qmark.cljc
```diff
@@ -0,0 +1,40 @@
...
+    (testing "uuids"
+      (are [x] (uuid? x)
+               #uuid "00000000-0000-0000-0000-000000000000"
+               #uuid "f81d4fae-7dec-11d0-a765-00a0c91e6bf6"
+               (parse-uuid "f81d4fae-7dec-11d0-a765-00a0c91e6bf6")
+               (random-uuid)))
+
+    (when-var-exists read-string
+      (testing "more uuids"
+        (read-string "#uuid \"f81d4fae-7dec-11d0-a765-00a0c91e6bf6\"")))
```
jeaye: This isn't testing `uuid?`, so I think we can remove it.

jeaye (changes requested):
Thanks for the PR!

PR #817, opened 2025-11-22
Add tests for `sort-by`

### LINKED ISSUE #493: `clojure.core/sort-by`

dgr:
Closes #493

### DIFF
--- a/test/clojure/core_test/sort_by.cljc
+++ b/test/clojure/core_test/sort_by.cljc
@@ -1,0 +1,151 @@
+(ns clojure.core-test.sort-by
+  (:require [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists]]))
+
+(when-var-exists sort-by
+  ;; Data for simple tests
+  (def simple-vec-maps [{:a 2 :b 9}
+                        {:a 1 :b 10}
+                        {:a 4 :b 7}
+                        {:a 3 :b 8}])
+
+  ;; Data to test stable sorting
+  (def stable-data [{:a 1 :b 11}
+                    {:a 1 :b 21}
+                    {:a 1 :b 39}
+                    {:a 1 :b 2}
+                    {:a 4 :b 5}
+                    {:a 2 :b 101}
+                    {:a 2 :b 71}
+                    {:a 2 :b 28}
+                    {:a 2 :b 99}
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/sort_by.cljc
```diff
@@ -0,0 +1,151 @@
...
+      (is (= (seq [{:a 4 :b 7}
+                   {:a 3 :b 8}
+                   {:a 2 :b 9}
+                   {:a 1 :b 10}])
+             (sort-by :b < simple-vec-maps)))
+      (is (= (seq [{:a 1 :b 10}
+                   {:a 2 :b 9}
+                   {:a 3 :b 8}
+                   {:a 4 :b 7}])
+             (sort-by :b > simple-vec-maps))))
```
jeaye: Are these testing any unique functionality that the compare + reverse compare cases don't cover? Looks to me like over-testing which we can remove.

jeaye (commented):
Thanks, Dave!

dgr:
I suspect the issue with the CLR tests is that `=` doesn't compare the `seq` and the return value from `sort-by` as equal, maybe because of a typing issue? I can't check CLR on my local Mac Air M1 as of now. Clojure returns a `clojure.lang.ArraySeq` from `sort-by`.

E-A-Griffin:
Yeah I can check tonight

dgr:
I just looked at things more closely. It's not related to comparison. For whatever reason, `sort-by` in CLR doesn't work with `<` or `>`, but seems to work with `compare`. @jeaye , see, those tests weren't redundant after all.

E-A-Griffin:
Tested locally and reproduced the same results

E-A-Griffin:
Need to verify but I think the issue is `sort` expects a comparator function that returns an integer value (C-style here similar to `strcmp` iirc) but the ClojureCLR instance is returning `true/false`

hence
```clj
user=> (sort-by :a #(if (> %1 %2) 1 (if (< %1 %2) -1 0)) simple-vec-maps)
({:a 1, :b 10} {:a 2, :b 9} {:a 3, :b 8} {:a 4, :b 7})
```
working as expected and also hence @dgr 's point about `compare` working

jeaye:
Nice digging, folks. If this is looking like a bug, we can ping David to loop him in.

dgr:
You just want me to add a trivial change of some sort?

dgr:
Tweaked. CI is running...

PR #820, opened 2025-11-23
Add tests for `<`

### LINKED ISSUE #119: `<`

dgr:
Closes #119

### DIFF
--- a/test/clojure/core_test/lt.cljc
+++ b/test/clojure/core_test/lt.cljc
@@ -92,6 +92,14 @@
            (is (= true (< "1" "2")))
            (is (= false (< "foo" "bar")))
            (is (= false (< :foo :bar)))]
+          :cljr
+          [(is (thrown? Exception (< nil 1)))
+           (is (thrown? Exception (< 1 nil)))
+           (is (thrown? Exception (< nil 1 2)))
+           (is (thrown? Exception (< 1 2 nil)))
+           (is (= true (< "1" "2")))
+           (is (thrown? Exception (< "foo" "bar")))
+           (is (thrown? Exception (< :foo :bar)))]
           :default
           [(is (thrown? Exception (< nil 1)))
            (is (thrown? Exception (< 1 nil)))


E-A-Griffin:
ClojureCLR seems to convert strings and chars to longs for inequality operators. I peeked at the ClojureCLR source code and messed around in the repl for a bit but I think this is worth tagging @dmiller 
```clj
$ cljr
Clojure core loaded in 176 milliseconds.
Clojure 1.12.2
user=> (< "1" "2")
true
user=> (< "2" "1")
false
user=> (< \1 \2)
true
user=> (< \2 \1)
false
user=> (< #"1" #"2")
Execution error (InvalidCastException) at System.Runtime.CompilerServices.CastHelpers/ChkCast_Helper (NO_FILE:0).
Unable to cast object of type 'System.Text.RegularExpressions.Regex' to type 'System.IConvertible'.
user=> (. clojure.lang.Numbers (lt "1" "2"))
true
```

dmiller:
This is an interesting problem.  The root cause is that Java has a `Number` class and the JVM does not.  I had to do a bunch of workarounds in `clojure.lang.Numbers` to deal with the mismatch.  And this results in an overgeneralization.    Where you have parameters of type `Number` on the JVM, I have to type them as `Object`.  That allows many things to slip into the internal machinery.  At some point they will get converted.  True numeric primitives are dealt with in their own manner.  For numeric types that are not primitive -- think of `BigDecimal` -- I call methods on the `Convert` class, such `Convert.ToInt64`.  For this, the value being converted must support the `System.IConvertible` interface.  `BigDecimal` does.  Unfortunately, so does `System.String`.  So if the string argument can be converted to a long, it will fly.   (The actual conversion are encoded in methods such as `clojure.lang.Util.ConvertoLong`.)

I coded this way because (1) I didn't have the `Number` type to help out, (2) it gave me the extension to `BigInteger` and `BigDecimal` (both of which were written by me), and didn't tie me to a fixed set of types -- I had no idea what else might get added in.    Unfortunately, it does over-generalize.  

The only solution to this over-generalization is to restrict the set of types considered.  This could be done by using more restricted versions of `Util.ConvertToLong` and kin to use in `Numbers`.  This could also be done by directly modifying the definitions of `Util.Convert.ToLong` and kin.   `Util.ConvertToLong` has something like 46 references.  And some of those are to functions used in core.  Either is a potentially breaking change, though to a coding practice that should not have been relied on.

What's the right solution?  Live with over-generalization?  Potentially break code?

dgr:
Done.

Side question: what's the best way to set up a CLJR environment for local testing on a Mac? Ideally, I could catch these CLR differences before I create PR and let CI run.

PR #822, opened 2025-11-24
Updates template to sort symbols in `:refer` clauses

dgr:
I noticed that @E-A-Griffin did a cleanup pass and was sorting the symbols in the `:refer` clauses of test namespaces. In order to promote future good hygiene, we should update the template to do what we prefer.


E-A-Griffin (approved):
Thanks! Thought I fixed this in that cleanup but must have missed it :)

E-A-Griffin:
Yeah I checked `main` and this does look stale

PR #790, opened 2025-11-07
Prep `double` for `jank`

shantanu-sardesai:
Created for the [port](https://github.com/jank-lang/jank/pull/583) in `jank`.

PR #828, opened 2025-11-25
Port arithmetic operators for jank

shantanu-sardesai:
Adding support for jank's [arithmetic promotion](https://github.com/jank-lang/jank/pull/588) operators.

### DIFF
--- a/test/clojure/core_test/plus_squote.cljc
+++ b/test/clojure/core_test/plus_squote.cljc
@@ -70,18 +70,19 @@
       6N 1N 5
       6N 1N 5N)
 
-    (is (thrown? #?(:cljs :default :clj Exception :cljr Exception) (+' 1 nil)))
-    (is (thrown? #?(:cljs :default :clj Exception :cljr Exception) (+' nil 1)))
+    #?(:jank []
+       :default [(is (thrown? #?(:cljs :default :clj Exception :cljr Exception) (+' 1 nil)))
+                 (is (thrown? #?(:cljs :default :clj Exception :cljr Exception) (+' nil 1)))])
 
-    (is (instance? clojure.lang.BigInt (+' 0 1N)))
-    (is (instance? clojure.lang.BigInt (+' 0N 1)))
-    (is (instance? clojure.lang.BigInt (+' 0N 1N)))
-    (is (instance? clojure.lang.BigInt (+' 1N 1)))
-    (is (instance? clojure.lang.BigInt (+' 1 1N)))
-    (is (instance? clojure.lang.BigInt (+' 1N 1N)))
-    (is (instance? clojure.lang.BigInt (+' 1 5N)))
-    (is (instance? clojure.lang.BigInt (+' 1N 5)))
-    (is (instance? clojure.lang.BigInt (+' 1N 5N)))
+    (is (#?@(:jank [cpp/jank.runtime.is_big_integer] :default [instance? clojure.lang.BigInt]) (+' 0 1N)))
+    (is (#?@(:jank [cpp/jank.runtime.is_big_integer] :default [instance? clojure.lang.BigInt]) (+' 0N 1)))
+    (is (#?@(:jank [cpp/jank.runtime.is_big_integer] :default [instance? clojure.lang.BigInt]) (+' 0N 1N)))
+    (is (#?@(:jank [cpp/jank.runtime.is_big_integer] :default [instance? clojure.lang.BigInt]) (+' 1N 1)))
+    (is (#?@(:jank [cpp/jank.runtime.is_big_integer] :default [instance? clojure.lang.BigInt]) (+' 1 1N)))
+    (is (#?@(:jank [cpp/jank.runtime.is_big_integer] :default [instance? clojure.lang.BigInt]) (+' 1N 1N)))
+    (is (#?@(:jank [cpp/jank.runtime.is_big_integer] :default [instance? clojure.lang.BigInt]) (+' 1 5N)))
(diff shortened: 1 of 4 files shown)


test/clojure/core_test/plus_squote.cljc
```diff
@@ -70,18 +70,19 @@
       6N 1N 5
       6N 1N 5N)
 
-    (is (thrown? #?(:cljs :default :clj Exception :cljr Exception) (+' 1 nil)))
-    (is (thrown? #?(:cljs :default :clj Exception :cljr Exception) (+' nil 1)))
+    #?(:jank []
```
shantanu-sardesai: Blocked by the https://github.com/jank-lang/jank/issues/558.

@jeaye

I was thinking, maybe it would be better to have a specific issue ticket in [jank](https://github.com/jank-lang/jank) (Maybe a child of [this ticket](https://github.com/jank-lang/jank/issues/415)) dedicated to updating all the Clojure test suite test cases for jank with support for catching specific exception types once the blocking feature is merged and in the meanwhile we merge these blocked PRs (this one, https://github.com/jank-lang/clojure-test-suite/pull/790, and more incomming). That way we have a reminder to update these test cases and benefit from the plethora of other test cases.

What do you think?

test/clojure/core_test/plus_squote.cljc
```diff
@@ -70,18 +70,19 @@
...
-    (is (instance? clojure.lang.BigInt (+' 0 1N)))
-    (is (instance? clojure.lang.BigInt (+' 0N 1)))
-    (is (instance? clojure.lang.BigInt (+' 0N 1N)))
-    (is (instance? clojure.lang.BigInt (+' 1N 1)))
-    (is (instance? clojure.lang.BigInt (+' 1 1N)))
-    (is (instance? clojure.lang.BigInt (+' 1N 1N)))
-    (is (instance? clojure.lang.BigInt (+' 1 5N)))
-    (is (instance? clojure.lang.BigInt (+' 1N 5)))
-    (is (instance? clojure.lang.BigInt (+' 1N 5N)))
+    (is (#?@(:jank [cpp/jank.runtime.is_big_integer] :default [instance? clojure.lang.BigInt]) (+' 0 1N)))
```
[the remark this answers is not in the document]
jeaye: I think repeating all of this over 10 times per file is not maintainable. Please use the `big-int?` predicate in `clojure.core-test.portability`. We can use `clojure.core-test.portability` in there.

test/clojure/core_test/plus_squote.cljc
shantanu-sardesai: Yeah, that's fine, so we can merge these PRs and just revisit these tests once jank can catch exceptions by type. Till then I can just keep a list of tests (in the comments of the ticket) that need to be updated.

test/clojure/core_test/plus_squote.cljc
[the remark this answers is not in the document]
shantanu-sardesai: Cool, we already have a utility. Done (for jank)!

test/clojure/core_test/portability.cljc
```diff
@@ -14,6 +14,7 @@
   ;; In CLJS, all numbers are really doubles and integer? and int?
   ;; return true if the fractional part of the double is zero
   #?(:cljs (integer? n)
+     :jank (cpp/jank.runtime.is_big_integer n)
      :default
      (and (integer? n)
           (not (int? n)))))
```
E-A-Griffin: If we're going to switch to this predicate, I think we could do a better job than this. Personally I think the default should be something like

```clj
(= (type 0N) (type n))
```

E-A-Griffin (changes requested):
Can we change the `:default` cases here to `Exception` since most platforms are going to throw an `Exception`? If you need to add an empty `:jank` case in the meantime we can always come back to it

jeaye:
> Can we change the `:default` cases here to `Exception` since most platforms are going to throw an `Exception`? If you need to add an empty `:jank` case in the meantime we can always come back to it

Agreed, this makes sense.

test/clojure/core_test/portability.cljc
dgr: I wrote the original default case for `big-int?`. I'm looking through the current version of it now and it seems like in some dialects it's becoming a synonym for `integer?`. I'm not sure that's what we want as it changes the semantics of it. IMO, there are a couple different cases that we need to think through. If a dialect doesn't act like Clojure JVM, there are a couple of choices. The dialect could hard fail on big integers, similar to what ClojureScript does on rationals. This would imply throwing an exception when reading a big integer (e.g., fail on "0N"). Or it could silently incorporate them for those in a given range. Thus, reading "0N" would succeed, but you'd have another kind of number, not a true big integer (again, similar to what JavaScript does). The question is what are the semantics of `big-int?` in the case of an implementation that just silently converts big integers to another type? In other words, is `big-int?` semantically equivalent to "Is this a big integer distinct from fixed-precision integers where `int?` returns `true`?" or is it semantically equivalent to "Is the type of this object the same type that is the result of `(read-string "0N")`?" Those are not quite the same thing. It feels like the `bit-int?` predicate semantics have gotten a little mushy. Specifically, the current code just accepts anything "big integerish" (and even "integerish") in a dialect that doesn't have big integers, and I'm not sure that's what we want. I'd have to go look at all the tests that use the `bit-int?` predicated to think this through more carefully, but just wanted to highlight the issue.

jeaye:
This has been inactive for all of 2026 so far, and it's quite conflicting, so I'm closing it. If you want to later reopen it, let me know.

PR #831, opened 2025-11-26
Redo `num`

E-A-Griffin:
Lots of cool finds from this! `num` is not a no-op in any dialect, but it's close to one in every dialect it's defined in. In `bb` it will throw for non numeric values, in `cljr` it never throws but does upcast for `int`, `short`, and `byte` (but interestingly not `float`). In `clj` it will box unboxed values which is very difficult to demonstrate, but not impossible, and like `bb` will throw for non-numeric values!

### DIFF
--- a/test/clojure/core_test/num.cljc
+++ b/test/clojure/core_test/num.cljc
@@ -27,114 +27,114 @@
 (when-var-exists num
  (deftest test-num
    (testing "positive cases"
-   #?@(:bb
-       [(is (NaN? (num ##NaN)))
-        (are [n] (and (= n (num n))
-                      (= (type n) (type (num n))))
-          0
-          0.1
-          1/2
-          1N
-          1.0M
-          (short 1)
-          (byte 1)
-          (int 1)
-          (long 1)
-          (float 1.0)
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -3,6 +3,9 @@
   :url "https://github.com/jank-lang/clojure-test-suite"
   :license {:name "MPL 2.0"
             :url "https://www.mozilla.org/en-US/MPL/2.0/"}
+  :clr {:main-cmd      ["Clojure.Main"]
+        :compile-cmd   ["Clojure.Compile"]}
   :dependencies [[org.clojure/clojure "1.12.0"]
                  [org.clojure/clojurescript "1.12.42"]]
-  :plugins [[com.jakemccrary/lein-test-refresh "0.25.0"]])
+  :plugins [[com.jakemccrary/lein-test-refresh "0.25.0"]
+            [lein-clr "0.2.2"]])
```
E-A-Griffin: Added these changes to make `cljr` development easier for myself

jeaye:
Thanks, Emma!

PR #835, opened 2025-11-29
Test defonce

### LINKED ISSUE #225: `clojure.core/defonce`

super7ramp:
This closes #225.

### DIFF
--- a/test/clojure/core_test/defonce.cljc
+++ b/test/clojure/core_test/defonce.cljc
@@ -5,11 +5,12 @@
 (when-var-exists defonce
   (deftest test-defonce
 
-    (testing "defining var only once"
-      (is (some? (defonce one 1)))
-      (is (nil? (defonce one 2)))
-      (is (nil? (defonce one (assert false "should not be evaluated"))))
-      (is (= 1 one)))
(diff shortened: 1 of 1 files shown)


super7ramp:
Might be related to https://github.com/thheller/shadow-cljs/issues/1185.

E-A-Griffin:
Could we test `^:private` metadata as well? You'll probably need a second namespace for this but would be good since it's undocumented behavior that `Clojure` supports so likely to differ between platforms

jeaye:
Thanks, @super7ramp! Apologies for the delay. I was on a 2.5 week road trip.

> Could we test `^:private` metadata as well? You'll probably need a second namespace for this but would be good since it's undocumented behavior that `Clojure` supports so likely to differ between platforms

What's the undocumented behavior here?

E-A-Griffin:
I *think* this is behavior for the full `def*` family of functions, special forms, and macros but I wouldn't be surprised if we uncover some inconsistencies between platforms in this realm, which we should test for

jeaye:
> I _think_ this is behavior for the full `def*` family of functions, special forms, and macros but I wouldn't be surprised if we uncover some inconsistencies between platforms in this realm, which we should test for

Hm. Do you know if the Clojure docs say this should happen anywhere? Some of the defined behavior comes from clojure.org instead of from the doc strings.

PR #851, opened 2026-02-22
Add tests for when-some, plus one for when-let.

alysbrooks:
`when-some` and `when-let` behave very similarly so this actually reuses a large number of tests from `when-let`. It also adds the opposite case of one of these tests to when-let.

### DIFF
--- a/test/clojure/core_test/when_let.cljc
+++ b/test/clojure/core_test/when_let.cljc
@@ -11,6 +11,8 @@
                      (is (nil? (when-let [x nil] x))))
                    (testing "basic single-binding tests using seqs"
                      (is (= '(0 1 2 3 4) (when-let [x (range 5)] x))))
+                   (testing "unlike when-some, we're looking for logical truth, so false doesn't evaluate"
+                     (is (nil? (when-let [x false] x))))
                    (testing "seq is only called once"
                      (let [calls (atom 0)
                            seq-fn (fn s [] (lazy-seq
--- a/test/clojure/core_test/when_some.cljc
+++ b/test/clojure/core_test/when_some.cljc
@@ -18,14 +18,6 @@
       (is (= '(0 1 2 3 4) (when-some [x (range 5)] x))))
     (testing "unlike when-let, we're looking for not-nil specifically, so false evaluates"
       (is (= false (when-some [x false] x))))
-    (testing "seq is only called once"
-      (let [calls (atom 0)
-            seq-fn (fn s [] (lazy-seq
-                              (swap! calls inc)
-                              (cons 1 (s))))
-            s (take 5 (seq-fn))]
-        (is (= '(1 1 1 1 1) (when-some [x s] x)))
-        (is (= @calls 5))))
     (testing "without a body, truth doesn't matter"
       (is (nil? (when-some [x nil])))
       (is (nil? (when-some [x [false]])))
--- a/test/clojure/core_test/when_some.cljc
+++ b/test/clojure/core_test/when_some.cljc
@@ -1,0 +1,32 @@
+
+(ns clojure.core-test.when-some
+  (:require [clojure.test :as t :refer [deftest is testing]]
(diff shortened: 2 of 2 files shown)


test/clojure/core_test/when_let.cljc
```diff
@@ -11,6 +11,8 @@
                      (is (nil? (when-let [x nil] x))))
                    (testing "basic single-binding tests using seqs"
                      (is (= '(0 1 2 3 4) (when-let [x (range 5)] x))))
+                   (testing "unlike, when-some, we're looking for not-nil specifically, so false evaluates"
```
E-A-Griffin: This is a good test but I feel like the testing string is backwards?

test/clojure/core_test/when_some.cljc
```diff
@@ -0,0 +1,32 @@
...
+    (testing "basic single-binding tests using seqs"
+      (is (= '(0 1 2 3 4) (when-some [x (range 5)] x))))
+    (testing "unlike when-let, we're looking for not-nil specifically, so false evaluates"
+      (is (= false (when-some [x false] x))))
+    (testing "seq is only called once"
+      (let [calls (atom 0)
+            seq-fn (fn s [] (lazy-seq
+                              (swap! calls inc)
+                              (cons 1 (s))))
+            s (take 5 (seq-fn))]
```
jeaye: This isn't testing that `seq` is called only once. Even if it's called twice, we're not necessarily going to re-run the whole lazy sequence. This is because when `seq` is called on a sequence, it just returns the sequence. `(take 5 (seq-fn))` gives us a sequence, so calling `seq` on it doesn't change anything.

I think this is an interesting property to try to test, but I don't know of a cross-dialect way to actually test it. @E-A-Griffin any ideas?

dgr:
What's the status of this PR? It's been two months since it was created and more than a month since the last comment.

alysbrooks:
Similar story as #852, it's been a hectic few months personally and professionally. I want to get back to it soon, though.

alysbrooks:
The last little thing was small enough—dropping the test that had an issue—so I was able to get to it tonight. Perhaps I'll have an idea on how to test the property some day.

E-A-Griffin (approved):

jeaye (approved):
LGTM!

test/clojure/core_test/when_some.cljc
```diff
@@ -0,0 +1,24 @@
...
+(when-var-exists when-some
+  (deftest test-when-some
+    (testing "basic single-binding tests using vectors or nil"
+      (is (= [0 1 2 3 4] (when-some [x [0 1 2 3 4]] x)))
+      (is (some? (when-some [x [nil]] x)))
+      (is (= [] (when-some [x []] x))))
+    (testing "side effects don't get evaluated"
+      (let [test-atom (atom nil)]
+        (is (nil? (when-some [x nil] (reset! test-atom 1) x)))
+        (is (nil? @test-atom))))
```
dgr, line 16: I'm not sure I understand what you're trying to get at with the atom and why returning a different, simpler, sentinel value wouldn't work. It seems overly complicated, but maybe I'm missing it. The body is just a `do` block, basically, so it either is evalutated or it isn't, and then value of the block is the final form.

test/clojure/core_test/when_some.cljc
```diff
@@ -0,0 +1,24 @@
...
+    (testing "basic single-binding tests using seqs"
+      (is (= '(0 1 2 3 4) (when-some [x (range 5)] x))))
```
dgr, line 18: Should also try this with an infinite range (`(range)`) to ensure that `when-some` doesn't force the sequence to be completely realized. Just return a single sentinel value in the body (e.g., `(first x)`) and test for that.

dgr (changes requested):

dgr:
Hi @alysbrooks , it has been a few months. What is your current thinking on this PR?

PR #861, opened 2026-04-15
Improve Babashka test tasks

dgr:
Does 3 things:
1. Prints banners when running each of the Babashka test tasks identifying the dialect. I think this helps identify which dialect was run when scrolling through CLI output.
2. Adds a `test-all` Babashka task which runs all of the test under all the dialects by calling the other tasks. Ideally, this should be run when adding new tests.
3. Removes (by way of commenting out) the `nrepl` Babashka task that was in the file. Is this being used??? I'm wondering if this was some copy/paste cruft because it's Babashka specific. I can delete it completely if wanted.

### DIFF
--- a/bb.edn
+++ b/bb.edn
@@ -1,21 +1,37 @@
 {:min-bb-version "1.12.210"
  :paths ["bb"]
  :tasks
- {test-jvm {:doc "Runs JVM tests"
-            :task (shell "lein test")}
+ {:init (defn banner
+         [message]
+         (let [stars (clojure.string/join (repeat 70 "*"))
+               padding-len (- 70 (count message) 5)
+               spaces (clojure.string/join (repeat padding-len " "))]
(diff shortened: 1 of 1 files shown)


jeaye:
Thanks, Dave! I'm good to remove that REPL task, since it doesn't affect CI.

dgr:
OK, go ahead and merge it as written for now and I'll submit another PR with that change plus another that I'm working on to give a "bb test --jvm --cljs --<dialect>" Babashka task using babashka/cli. That helps people that don't want to run everything easily run all the test suites they want in a single command. I'm working on that now.

jeaye:
Nice! Maybe we don't even need the `--` before those. `bb test jvm cljs` seems like it'd be enough.

dgr:
I could do it that way, too. Let me fiddle.

PR #865, opened 2026-04-17
Add tests for `>=`

### LINKED ISSUE #123: `>=`

dgr:
Closes #123

### DIFF
--- a/test/clojure/core_test/gt_eq.cljc
+++ b/test/clojure/core_test/gt_eq.cljc
@@ -77,6 +77,7 @@
         true 2 1 0
         true 0 -1 -2
         true 1 0 -1
+        true 1 1 1
         false 2 0 1
         false 0 2 1
         false -1 -2 0
@@ -89,7 +90,8 @@
         false ##-Inf 0 ##Inf)
       (is (= true (apply >= (reverse (range 10)))))
       (is (= true (apply >= 10 (reverse (range 10)))))
-      (is (= false (apply >= -1 (reverse (range 10))))))
+      (is (= false (apply >= -1 (reverse (range 10)))))
+      (is (= true (apply >= (repeat 5 1)))))
 
     (testing "negative tests"
       ;; `<=` only compares numbers, except in ClojureScript (really
--- a/test/clojure/core_test/gt_eq.cljc
+++ b/test/clojure/core_test/gt_eq.cljc
@@ -5,15 +5,16 @@
 (when-var-exists >=
   (deftest test->=
     (testing "arity 1"
-      ;; Doesn't matter what the argument is, `>=` return `true` for
-      ;; one argument.
-      (is (>= 1))
-      (is (>= 0))
-      (is (>= -1))
-      ;; Doesn't check whether arg is a number
-      (is (>= "abc"))
-      (is (>= :foo))
-      (is (>= nil)))
+      (are [x] (= true (>= x))
+        ;; Doesn't matter what the argument is, `>=` return `true` for
+        ;; one argument.
+        1
+        0
+        -1
+        ;; Doesn't check whether the argument is a number
+        "abc"
+        :foo
+        nil))
 
     (testing "arity 2"
       (are [expected x y] (= expected (>= x y))
--- a/test/clojure/core_test/gt_eq.cljc
+++ b/test/clojure/core_test/gt_eq.cljc
@@ -42,10 +42,10 @@
         true ##Inf 1
 
         true 0 0
-        true  1 1
-        true  -1 -1
-        true  ##Inf ##Inf
-        true  ##-Inf ##-Inf
+        true 1 1
+        true -1 -1
+        true ##Inf ##Inf
+        true ##-Inf ##-Inf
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/gt_eq.cljc
```diff
@@ -0,0 +1,128 @@
+(ns clojure.core-test.gt-eq
+  (:require [clojure.test :as t :refer [are deftest is testing]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists] :as p]))
+
+(when-var-exists >=
+  (deftest test->=
```
E-A-Griffin, line 6: nitpick, I think `test>=` is more clear since to me this looks like we're testing some function/macro `->=`

test/clojure/core_test/gt_eq.cljc
```diff
@@ -0,0 +1,128 @@
...
+        true 0.0M -1.0M
+        true -1 ##-Inf
+        true ##Inf 1
+
+        true 0 0
+        true  1 1
+        true  -1 -1
+        true  ##Inf ##Inf
+        true  ##-Inf ##-Inf
+        false ##NaN ##NaN               ; ##NaN is never equal, even to itself
```
E-A-Griffin, line 50: indentation is weird here

test/clojure/core_test/gt_eq.cljc
```diff
@@ -0,0 +1,128 @@
...
+        true 1 1 0
+        false 0 0 1
+        false 0 1 1
+        true ##Inf 0 ##-Inf
+        false ##-Inf 0 ##Inf)
+      (is (= true (apply >= (reverse (range 10)))))
+      (is (= true (apply >= 10 (reverse (range 10)))))
+      (is (= false (apply >= -1 (reverse (range 10))))))
+
+    (testing "negative tests"
```
jeaye, line 96: I think we'll want to scale back on these negative tests, since they're technically undefined behavior in Clojure. The docs string for `>=` says `Returns non-nil if nums are in monotonically non-decreasing order, otherwise false.` so there is an implication that all inputs must be `nums`.

Interestingly, the doc string also says `non-nil`, rather than `true`.

E-A-Griffin (approved):
Small things but looks good to me!

test/clojure/core_test/gt_eq.cljc
E-A-Griffin: Yeah "returns `non-nil`" is very oddly non specific

E-A-Griffin:
One thing that I think could be a good add would be binary numeric comparisons for applicable hosts, Clojure, ClojureScript, ClojureCLR, bb for example have
```
user=> (>= 2r0110 2r0100)
true
```

E-A-Griffin:
Also would be good to test negative Infinity

jeaye:
> One thing that I think could be a good add would be binary numeric comparisons for applicable hosts, Clojure, ClojureScript, ClojureCLR, bb for example have
> 
> ```
> user=> (>= 2r0110 2r0100)
> true
> ```

I like the out-of-the-box thinking, but in this case, `2r0110` is just parsed into a normal number and the behavior of `>=` is not different. We'd just be testing the parser, which is not the goal here.

jeaye:
> Also would be good to test negative Infinity

Good thinking.

test/clojure/core_test/gt_eq.cljc
dgr: The template has a standard format: "test-" appended with function name. I'm going to leave this as is unless you have a lot of heartburn, just for consistency with every other test.

test/clojure/core_test/gt_eq.cljc
dgr: Fixed.

test/clojure/core_test/gt_eq.cljc
dgr: I'll switch the tests to check for non-nil and `false` as opposed to `true` and `false`.

dgr: Also removed all negative tests for anything other than numbers and `nil`.

dgr:
> Also would be good to test negative Infinity

There are multiple tests with `##-Inf`, so I'm not sure what you're looking for here.

dgr:
> > One thing that I think could be a good add would be binary numeric comparisons for applicable hosts, Clojure, ClojureScript, ClojureCLR, bb for example have
> > ```
> > user=> (>= 2r0110 2r0100)
> > true
> > ```
> 
> I like the out-of-the-box thinking, but in this case, `2r0110` is just parsed into a normal number and the behavior of `>=` is not different. We'd just be testing the parser, which is not the goal here.

I agree with this. There's no added benefit as the syntax is just specific to the reader and it's all converted to integers by the time it reaches `>=`.

dgr:
Changing this to draft PR until I can change the tests to check for non-nil rather than `true`.

dgr:
Alright, I think this is ready to go now.

jeaye:
Thanks, Dave!

PR #868, opened 2026-04-18
Add tests for `eval`

### LINKED ISSUE #259: `eval`

dgr:
Closes #259

### DIFF
--- a/test/clojure/core_test/eval.cljc
+++ b/test/clojure/core_test/eval.cljc
@@ -8,45 +8,46 @@
 
   (deftest test-eval
     #?(:cljs nil
-       :default (testing "Strings, numbers, characters, true, false, nil and keywords evaluate to themselves."
-                  (are [expected form] (= expected (eval form))
-                    ;; lots of Clojure objects just evaluate to themselves
-                    1 1
-                    0 0
-                    -1 -1
-                    1.0 1.0
-                    1N 1N
-                    1.0M 1.0M
-                    1/2 1/2
-                    "a string" "a string"
-                    "(+ 1 2)" "(+ 1 2)" ; strings are just evaluated as strings
-                    \x \x
-                    true true
-                    false false
-                    nil nil
-                    :a-keyword :a-keyword)))
-    #?(:cljs nil
-       :default (testing "Symbol resolution"
-                  ;; namespace qualified
-                  (is (= 42 (eval 'clojure.core-test.eval/x)))))
-    #?(:cljs nil
-       :default (testing "Vectors, Maps, Sets"
-                  ;; basic literal collections
-                  (is (= [:a :b] (eval [:a :b])))
-                  (is (= {:a :b} (eval {:a :b})))
-                  (is (= #{:a :b} (eval #{:a :b})))
-                  ;; collections with embedded symbols
-                  (is (= [:a :b 42] (eval [:a :b 'clojure.core-test.eval/x])))))
-    #?(:cljs nil
-       :default (testing "Lists, function application, and macros"
-                  ;; empty list evaluates to itself
-                  (is (= '() (eval '())))
-                  ;; function calls
-                  (is (= 5 (eval '(+ 2 3))))
-                  (is (= 6 (eval '(* 2 3))))
-                  ;; macros
-                  (is (= 42 (eval '(or false clojure.core-test.eval/x))))
-                  (is (= 42 (eval '(and (+ 2 3) clojure.core-test.eval/x))))
-                  ;; special forms
-                  (is (= 43 (eval '(let [y 43] (or false y)))))
-                  (is (= 43 (eval '(loop [y 0] (if (= y 43) y (recur (inc y)))))))))))
+       :default
+       (do (testing "Strings, numbers, characters, true, false, nil and keywords evaluate to themselves."
+             (are [expected form] (= expected (eval form))
+               ;; lots of Clojure objects just evaluate to themselves
+               1 1
+               0 0
+               -1 -1
+               1.0 1.0
+               1N 1N
+               1.0M 1.0M
+               1/2 1/2
+               "a string" "a string"
+               "(+ 1 2)" "(+ 1 2)" ; strings are just evaluated as strings
+               \x \x
+               true true
+               false false
+               nil nil
+               :a-keyword :a-keyword))
+
+           (testing "Symbol resolution"
+             ;; namespace qualified
+             (is (= 42 (eval 'clojure.core-test.eval/x))))
+
+           (testing "Vectors, Maps, Sets"
+             ;; basic literal collections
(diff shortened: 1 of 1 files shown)


dgr:
Note that this does not test ClojureScript `eval`, but I need some feedback on the way that I handled the reader conditionalization. Specifically, ClojureScript still includes an `eval` symbol, so the `when-var-exists` macro will find it and not print `SKIP - eval`. Thus, the tests, if any, in the namespace will be executed. I could have conditionalized the whole `deftest` form but if I had done that, then there would have been no test reported in the ClojureScript output. So, instead, I conditionalized each of the `testing` blocks inside the `deftest` form. Thus, ClojureScript sees a `deftest` that contains no assertions. @jeaye, if you want me to do it another way, I'm happy to adjust it.

jeaye:
> Thus, ClojureScript sees a deftest that contains no assertions. @jeaye, if you want me to do it another way, I'm happy to adjust it.

This is how I'd do it, too. All good.

jeaye:
Oh, actually, can we just have one reader conditional, as a top-level `do` which is `nil` for CLJS?

test/clojure/core_test/eval.cljc
```diff
@@ -0,0 +1,52 @@
...
+                  (is (= '() (eval '())))
+                  ;; function calls
+                  (is (= 5 (eval '(+ 2 3))))
+                  (is (= 6 (eval '(* 2 3))))
+                  ;; macros
+                  (is (= 42 (eval '(or false clojure.core-test.eval/x))))
+                  (is (= 42 (eval '(and (+ 2 3) clojure.core-test.eval/x))))
+                  ;; special forms
+                  (is (= 43 (eval '(let [y 43] (or false y)))))
+                  (is (= 43 (eval '(loop [y 0] (if (= y 43) y (recur (inc y)))))))))))
```
jeaye: Note that both of these are macros and not special forms. That's not to say this is wrong, it's just not exactly testing what's commented. `let*` and `loop*` are the special forms, which would work just as well here. The macros just add destructuring support.

test/clojure/core_test/eval.cljc
```diff
@@ -0,0 +1,52 @@
...
+                    1N 1N
+                    1.0M 1.0M
+                    1/2 1/2
+                    "a string" "a string"
+                    "(+ 1 2)" "(+ 1 2)" ; strings are just evaluated as strings
+                    \x \x
+                    true true
+                    false false
+                    nil nil
+                    :a-keyword :a-keyword)))
```
jeaye: What about eval for a fn object?

test/clojure/core_test/eval.cljc
```diff
@@ -0,0 +1,52 @@
...
+                  ;; basic literal collections
+                  (is (= [:a :b] (eval [:a :b])))
+                  (is (= {:a :b} (eval {:a :b})))
+                  (is (= #{:a :b} (eval #{:a :b})))
+                  ;; collections with embedded symbols
+                  (is (= [:a :b 42] (eval [:a :b 'clojure.core-test.eval/x])))))
+    #?(:cljs nil
+       :default (testing "Lists, function application, and macros"
+                  ;; empty list evaluates to itself
+                  (is (= '() (eval '())))
```
jeaye: What about an infinite seq?

jeaye (commented):
Thanks, Dave!

test/clojure/core_test/eval.cljc
dgr: Hm. I relied on https://clojure.org/reference/special_forms for the list of special forms. And the doc strings for both `let` and `loop` say they are special forms. For example:

```
user> (doc loop)
-------------------------
clojure.core/loop
  (loop [bindings*] exprs*)
([bindings & body])
Special Form
  Evaluates the exprs in a lexical context in which the symbols in
  the binding-forms are bound to their respective init-exprs or parts
  therein. Acts as a recur target.

  Please see http://clojure.org/special_forms#loop
```

and

```
user> (doc let)
-------------------------
clojure.core/let
  (let [bindings*] exprs*)
([bindings & body])
Special Form
  binding => binding-form init-expr
  binding-form => name, or destructuring-form
  destructuring-form => map-destructure-form, or seq-destructure-form

  Evaluates the exprs in a lexical context in which the symbols in
  the binding-forms are bound to their respective init-exprs or parts
  therein.

  See https://clojure.org/reference/special_forms#binding-forms for
  more information about destructuring.

  Please see http://clojure.org/special_forms#let
Spec
  args: (cat :bindings :clojure.core.specs.alpha/bindings :body (* any?))
  ret: any?
```

test/clojure/core_test/eval.cljc
dgr: Yes, good call. I will add.

test/clojure/core_test/eval.cljc
dgr: What are you trying to test with that? That it doesn't get realized?

test/clojure/core_test/eval.cljc
dgr: Added.

test/clojure/core_test/eval.cljc
dgr: So, eval'ing an infinite range returns a `clojure.lang.Iterate` object. I'm not even sure what that is.

```
user> (class (eval '(range)))
clojure.lang.Iterate
```

What are you wanting to test with that?

dgr: Interestingly, eval'ing a fixed range returns a `clojure.lang.LongRange`:

```
user> (class (eval '(range 10)))
clojure.lang.LongRange
```

jeaye: > What are you trying to test with that? That it doesn't get realized?

Yep.

> So, eval'ing an infinite range returns a clojure.lang.Iterate object. I'm not even sure what that is.

That's the type of `(range)`.

> Interestingly, eval'ing a fixed range returns a clojure.lang.LongRange:

`LongRange` was the type that went in. The `Long` is because each value is a `Long` (i.e. `10` is a `Long`).

So, eval clearly doesn't realize the seq, from the behavior. That's a great test case to have.

dgr: So, how do I test for not being realized, exactly? Is it just that I get something back at all, because if it tried to realize the infinite sequence it would wait until the full sequence was realized and finally throw and OOM exception?

jeaye: Exactly. If we get it back and continue running the test, we can know that it wasn't realized.

dgr: OK, I'll add this.

dgr:
I think this is good to go now.

jeaye:
Great work, Dave!

PR #870, opened 2026-04-21
Add tests for `apply`

dgr:


test/clojure/core_test/apply.cljc
```diff
@@ -0,0 +1,41 @@
...
+    (is (= 1 (apply + 1 {})))
+    (is (= 1 (apply + 1 #{})))
+    (is (= 1 (apply + 1 "")))
+    
+    (is (= 1 (apply + '(1))))           ; Zero non-sequence arg
+    (is (= 3 (apply + 1 '(2))))         ; One non-sequential arg
+    (is (= 6 (apply + 1 2 '(3))))       ; Two non-sequential args
+    (is (= 10 (apply + 1 2 3 '(4))))    ; Three non-sequential args
+    (is (= 15 (apply + 1 2 3 4 '(5))))  ; Four non-sequential args
+    (is (= 45 (apply + (range 10))))    ; All args as sequential
```
jeaye, line 26: What do you think of a test where an arg is an infinite seq?

```clojure
(is (= 3 (count (apply conj [] [1 2 (range)])))
```

test/clojure/core_test/apply.cljc
```diff
@@ -0,0 +1,41 @@
...
+    (is (= 0 (apply + "")))             ; string is a sequence of characters
+    
+    (is (= 1 (apply + 1 nil)))          ; Apply + with 1 arg and empty seqeuence
+    (is (= 1 (apply + 1 '())))
+    (is (= 1 (apply + 1 [])))
+    (is (= 1 (apply + 1 {})))
+    (is (= 1 (apply + 1 #{})))
+    (is (= 1 (apply + 1 "")))
+    
+    (is (= 1 (apply + '(1))))           ; Zero non-sequence arg
```
jeaye, line 21: Maybe interesting?

```clojure
(is (= 10 (apply apply + [1 2 [3 4]]))
```

test/clojure/core_test/apply.cljc
NoahTheDuke: I believe that does or did trip up [Glojure](https://github.com/gloathub/glojure), so it's a good test case.

dgr: I thought about that but decided against it as over-testing. That test belongs in `eval` since that's where it might get realized or not. In `apply`, we're just collecting all the function arguments and then passing them to the function being invoked. So, the args are just opaque things to `apply`.

test/clojure/core_test/apply.cljc
dgr: Yes, that's interesting. I did a recursive `eval` in that set of tests. I should do the same here. Good catch.

test/clojure/core_test/apply.cljc
jeaye: > In apply, we're just collecting all the function arguments and then passing them to the function being invoked. So, the args are just opaque things to apply.

That's the assumption of a well-behaving dialect, which is exactly the thing we're trying to test here. Is the dialect well-behaving? There's only one way to know. :)

jeaye: And if it tripped up Glojure already, that's even more reason to add it.

E-A-Griffin:
`apply` with an inequality function (e.g. `<`) could be helpful given the wonkiness we've caught with these prior and how different these functions operate than their operator equivalents in most PLs

dgr:
> `apply` with an inequality function (e.g. `<`) could be helpful given the wonkiness we've caught with these prior and how different these functions operate than their operator equivalents in most PLs

What specifically are you trying to test? `apply` is pretty simple. It basically just validates that the first argument is an `IFn` and calls the correct arity with the supplied parameters. Given that, what do you think could go wrong with `<` that isn't verified by the test assertions in the current PR?

jeaye:
> > `apply` with an inequality function (e.g. `<`) could be helpful given the wonkiness we've caught with these prior and how different these functions operate than their operator equivalents in most PLs
> 
> What specifically are you trying to test? `apply` is pretty simple. It basically just validates that the first argument is an `IFn` and calls the correct arity with the supplied parameters. Given that, what do you think could go wrong with `<` that isn't verified by the test assertions in the current PR?

Thanks for the idea, Emma! I agree with Dave here that this one shouldn't be needed. Keep 'em coming.

dgr:
@jeaye , BTW, good idea to invoke the CoPilot review. I should do that more often. It caught a couple formatting things (trailing white space, comment on the wrong line, and a misspelling. I guess a little AI isn't a bad thing. :-)

jeaye:
> @jeaye , BTW, good idea to invoke the CoPilot review. I should do that more often. It caught a couple formatting things (trailing white space, comment on the wrong line, and a misspelling. I guess a little AI isn't a bad thing. :-)

I'm glad you're not annoyed by it! I've been using it for my jank PRs and it actually catches a lot. So far, it's definitely been worth my while. It's free for the jank-lang org, so, as you said, we might as well.

test/clojure/core_test/apply.cljc
dgr: OK, I can add it. Simple enough.

dgr: Done.

jeaye (approved):
LGTM. Awaiting approval from Emma, as she may have more comments.

dgr:
OK, this should be ready now.

dgr:
BTW, @E-A-Griffin , there are some tests of `apply` with comparison functions in `gt_eq.cljc` and `lt_eq.cljc`, so it's not completely untested.

E-A-Griffin (approved):
Let's get this merged!

PR #876, opened 2026-05-09
Switch Basilisp test suite to use only 2 CPUs

chrisrink10:
`pytest-xdist`'s `auto` option can often be slower than setting a specific number of CPUs. I think standard Github Actions workers only have 2 vCPUs so this should end up actually being faster than using `auto`.

I also bumped the different Github Actions versions to use newer releases which use Node 24 to suppress some warnings during the runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,10 +14,10 @@ jobs:
   test-jvm:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
       - name: Prepare java
-        uses: actions/setup-java@v4
+        uses: actions/setup-java@v5
         with:
           distribution: "temurin"
           java-version: "21"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -127,7 +127,7 @@ jobs:
   test-clr:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
       - name: Prepare dotnet
         uses: xt0rted/setup-dotnet@v1.5.0
```
chrisrink10, line 133: I'm not sure why this action was selected, but it appears to be out of sync from the [official setup-dotnet](https://github.com/actions/setup-dotnet) action from Github and hasn't been updated since 2022.

jeaye: @dmiller Any reason to not use the official `setup-dotnet`?

jeaye:
Thanks for the speedup, Chris!

PR #878, opened 2026-05-13
Add tests for `repeatedly`

### LINKED ISSUE #456: `clojure.core/repeatedly`

E-A-Griffin:
Closes https://github.com/jank-lang/clojure-test-suite/issues/456


test/clojure/core_test/repeatedly.cljc
```diff
@@ -0,0 +1,65 @@
...
+        (is (= 5 @state)))
+      (testing "handles mid failures gracefully"
+        (let [state (atom 0)
+              fails-second-run (fn [n] (if (> @n 0)
+                                        (throw (ex-info "expected" {}))
+                                        (swap! n inc)))]
+          (try
+            (last (repeatedly 2 #(fails-second-run state)))
+            (catch #?(:cljs    :default
+                      :default Exception) _ nil))
```
jeaye: This isn't going to work for some dialects, like jank. We have `p/thrown?` to handle exception details. Perhaps we can rework this test to use `p/thrown?`. If not, I think we should add a `(p/exception)` or similar, to get the base exception type for any given dialect.

test/clojure/core_test/repeatedly.cljc
```diff
@@ -0,0 +1,65 @@
...
+      (testing "zero returns empty list"
+        (is (= '() (repeatedly 0 +))))
+      (testing "natural numbers"
+        (is (= '(0) (repeatedly 1 +)))
+        (is (= '(0 0 0) (repeatedly 3 +))))
+      (testing "non-integer numbers"
+        (is (= '(0 0) (repeatedly 1.5 +)))
+        (is (= '(0)
+               #?(:cljs '(0) ;; cljs doesn't implement ratios
+                  :default (repeatedly 1/2 +))))
```
jeaye: CLJS supports reading `1/2` into `0.5`, so maybe we just keep this sort of test either way. If CLJS later adds ratio support, we'll already have it covered.

test/clojure/core_test/repeatedly.cljc
```diff
@@ -0,0 +1,65 @@
...
+                                        (swap! n inc)))]
+          (try
+            (last (repeatedly 2 #(fails-second-run state)))
+            (catch #?(:cljs    :default
+                      :default Exception) _ nil))
+          (is (= @state 1))))
+      (testing "is lazy"
+        (let [state (atom 0)]
+          #_{:clj-kondo/ignore [:redundant-do]}
+          (do (repeatedly #(swap! state inc))
```
jeaye: Why not just put this guy in the `let`? No need for the kondo bit or the `do`.

jeaye (changes requested):
Thanks for the help, Emma!

test/clojure/core_test/repeatedly.cljc
E-A-Griffin: Yeah I think we should probably implement `p/exception` anyways, and I can make and take a cleanup issue for it

test/clojure/core_test/repeatedly.cljc
E-A-Griffin: Yeah I can do that!

test/clojure/core_test/repeatedly.cljc
dgr: Yea, exceptions are difficult to deal with in Clojure dialects. This is one area where dialects should maybe try to come together, otherwise portable code is pretty much impossible. You basically just have to avoid anything throwing.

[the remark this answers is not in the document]
dgr: Wow, I'm impressed that CoPilot actually caught this.

test/clojure/core_test/repeatedly.cljc
```diff
@@ -0,0 +1,65 @@
...
+              (is (= 0 @state))))))
+
+    (testing "Single argument"
+      (is (= 0 (first (repeatedly +)))))
```
dgr: You might want to test the remainder of the seq for laziness.

dgr (approved):
Overall, looks good.

test/clojure/core_test/repeatedly.cljc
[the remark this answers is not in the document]
E-A-Griffin: yeah me too

test/clojure/core_test/repeatedly.cljc
E-A-Griffin: I tried this and it broke tests, can you confirm that ratios work on your end and I can look into my setup?

jeaye: Hm, I just went to https://cljs.io/ and tried `1/3` and it gave `0.333333`. I then cross-referenced that with the [differences from Clojure](https://www.clojurescript.org/about/differences) doc which says `Ratio` isn't supported.

So I used this to see that CLJS supports reading ratios, but doesn't have a ratio object and instead just reads them into floats.

E-A-Griffin:
> Hm, I just went to https://cljs.io/ and tried `1/3` and it gave `0.333333`. I then cross-referenced that with the [differences from Clojure](https://www.clojurescript.org/about/differences) doc which says `Ratio` isn't supported.
> 
> So I used this to see that CLJS supports reading ratios, but doesn't have a ratio object and instead just reads them into floats.

When I ran it on the test suite locally I got an exception so I think the local CLJS we're using doesn't support it

jeaye:
> > Hm, I just went to https://cljs.io/ and tried `1/3` and it gave `0.333333`. I then cross-referenced that with the [differences from Clojure](https://www.clojurescript.org/about/differences) doc which says `Ratio` isn't supported.
> > So I used this to see that CLJS supports reading ratios, but doesn't have a ratio object and instead just reads them into floats.
> 
> When I ran it on the test suite locally I got an exception so I think the local CLJS we're using doesn't support it

Okie dokie! Let's not worry about it for now then.

dgr:
> > > Hm, I just went to https://cljs.io/ and tried `1/3` and it gave `0.333333`. I then cross-referenced that with the [differences from Clojure](https://www.clojurescript.org/about/differences) doc which says `Ratio` isn't supported.
> > > So I used this to see that CLJS supports reading ratios, but doesn't have a ratio object and instead just reads them into floats.
> > 
> > 
> > When I ran it on the test suite locally I got an exception so I think the local CLJS we're using doesn't support it
> 
> Okie dokie! Let's not worry about it for now then.

Maybe shoot a quick Slack message out to #cljs-dev.

test/clojure/core_test/repeatedly.cljc
E-A-Griffin: How would you recommend doing this?

dgr: I think you generate an infinite seq (`(range)`) and then realize the first element and test that. If it doesn't throw, you're good to go. If it realizes the whole thing, then you'll get an OOM exception.

dgr: Should also test that `repeatedly` returns an un-`realized?` `p/lazy-seq?`.

dgr (changes requested):

jeaye (approved):
Nice work, Emma!

PR #883, opened 2026-05-16
Add tests for `lazy-seq`

### LINKED ISSUE #332: `clojure.core/lazy-seq`

dgr:
Includes a `lazy-seq?` predicate in portability.cljc.

Closes #332

### DIFF
--- a/test/clojure/core_test/lazy_seq.cljc
+++ b/test/clojure/core_test/lazy_seq.cljc
@@ -1,0 +1,72 @@
+(ns clojure.core-test.lazy-seq
+  (:require [clojure.test :as t :refer [are deftest is testing]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists] :as p]))
+
+(when-var-exists lazy-seq
+
+  ;; Note: we cannot use Clojure's standard `range` here because
+  ;; Clojure JVM uses a `clojure.lang.LongRange` which is chunked and
+  ;; is not a lazy seq, in spite of what the `range` doc string says.
+  (defn lazy-infinite-range
+    ([] (lazy-infinite-range 0))
+    ([n] (lazy-seq (cons n (lazy-infinite-range (inc n))))))
+
+  (defn nthrest*
+    "The standard version of `nthrest` calls `seq` at a certain point and
+  forces lazy seqs to become partially realized. This custom version
+  only calls `rest` on the collection the number of times specified by
+  `n`."
+    [coll n]
+    (assert (not (neg? n)))
+    (loop [n n
+           coll coll]
+      (if (zero? n)
+        coll
+        (recur (dec n) (rest coll)))))
+  
+  (deftest test-lazy-seq
+    ;; Note: lazy seqs are stateful. Once a lazy seq is realized, it
+    ;; cannot be unrealized. Holding a reference to the lazy seq from
+    ;; when it was unrealized doesn't prevent the reference from
+    ;; seeing its realized state.
+    (testing "basics"
+      ;; Super short lazy seq
+      (let [s (lazy-seq (cons 1 nil))]
+        (is (p/lazy-seq? s))
+        (is (not (realized? s)))
+        (is (= [1] (doall s)))
+        (is (realized? s)))
+
+      ;; The realized value of a lazy seq just has to be a seq, not
+      ;; necessarily a cons, so we can return a whole vector.
+      (let [s (lazy-seq [1 2 3])]
+        (is (p/lazy-seq? s))
+        (is (not (realized? s)))
+        (is (= [1 2 3] (doall s)))
+        (is (realized? s)))
+
+      (let [s (lazy-infinite-range)]
+        (is (p/lazy-seq? s))
+        (is (not (realized? s)))
+        (is (= 0 (first s)))
+        (is (realized? s))               ; Calling `first` realizes lazy seq
+        (is (p/lazy-seq? (rest s)))      ; but the rest remains lazy
+        (is (not (realized? (rest s))))  ; and unrealized.
+        (is (p/lazy-seq? (nthrest* s 50))) ; nthrest* forces realization of everything before
+        (is (not (realized? (nthrest* s 50))))
+        (is (p/lazy-seq? (nthrest* s 49)))
+        (is (realized? (nthrest* s 49))))
+
+      ;; Remember `(next coll)` = `(seq (rest coll))` and `seq`
+      ;; realizes the element and returns a seq rather than a lazy
+      ;; seq.
+      (let [s (lazy-infinite-range)]
+        (is (not (p/lazy-seq? (next s)))) ; `next` returns seq not lazy seq
+        (is (realized? (rest s))) ; `next` forces realization of next element
+        (is (p/lazy-seq? (nthrest* s 10)))
+        (is (not (realized? (nthrest* s 10))))))  ; but further elements not realized
+    (testing "negative-cases"
+      ;; The realized value of `lazy-seq` must be a seq (i.e., satisfy
+      ;; `ISeq` or its dialect-specific equivalent)
+      (is (p/thrown? (first (lazy-seq 1))))) ; a long is not a seq
+    ))
--- a/test/clojure/core_test/lazy_seq.cljc
+++ b/test/clojure/core_test/lazy_seq.cljc
@@ -61,6 +61,7 @@
       ;; realizes the element and returns a seq rather than a lazy
       ;; seq.
       (let [s (lazy-infinite-range)]
+        (next s)
         (is (realized? (rest s))) ; `next` forces realization of next element
         (is (p/lazy-seq? (nthrest* s 10)))
         (is (not (realized? (nthrest* s 10))))))  ; but further elements not realized
--- a/test/clojure/core_test/lazy_seq.cljc
+++ b/test/clojure/core_test/lazy_seq.cljc
@@ -71,7 +71,17 @@
       ;; the lazy seq is realized, Clojure will keep realizing lazy
       ;; seqs until it bottoms out with something that is not a lazy
       ;; seq
-      (let [s (lazy-seq (lazy-seq (lazy-seq (cons 1 '(2 3)))))]
+      (let [s (lazy-seq
+               (lazy-seq
+                (lazy-seq
+                 (cons 1 (lazy-seq
+                          (lazy-seq
+                           (lazy-seq
+                            (lazy-seq (cons 2 (lazy-seq
+                                               (lazy-seq
+                                                (lazy-seq
+                                                 (lazy-seq
+                                                  (lazy-seq (cons 3 nil)))))))))))))))]
         (is (= 1 (first s)))
         (is (= 2 (first (next s))))
         (is (= 3 (first (next (next s))))))
--- a/test/clojure/core_test/lazy_seq.cljc
+++ b/test/clojure/core_test/lazy_seq.cljc
@@ -23,7 +23,7 @@
       (if (zero? n)
         coll
         (recur (dec n) (rest coll)))))
-  
+
   (deftest test-lazy-seq
     ;; Note: lazy seqs are stateful. Once a lazy seq is realized, it
     ;; cannot be unrealized. Holding a reference to the lazy seq from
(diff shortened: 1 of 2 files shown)


dgr:
I know this will blow up in Basilisp and Phel. Need to add dialect-specific code to `lazy-seq?`.

dgr:
@jasalt, @Chemaclass, and @chrisrink10  if you can tell me the right code for fixing `lazy-seq?` for your dialects, I'll add it.

Chemaclass:
@dgr for Phel:                                                                                                                                                
```clojure
  :phel (phel.core/lazy-seq? x)                                                                                                                                   
```                                                                                                                                                     
Needs Phel `^0.38` (predicate added in https://github.com/phel-lang/phel-lang/pull/1992). Bump PR: #884 rebase on main once merged.  Namespace-qualify to avoid self-recursion against the surrounding defn.

Chemaclass:
Filed the Phel-side semantic gaps upstream — both reproduced on 0.38:

- [phel-lang/phel-lang#1994](https://github.com/phel-lang/phel-lang/issues/1994) — `next` on a lazy-seq returns a `LazySeq` instead of a realized seq → breaks `(not (p/lazy-seq? (next s)))`.
- [phel-lang/phel-lang#1995](https://github.com/phel-lang/phel-lang/issues/1995) — `LazySeq::cdr` eagerly realizes the tail head → breaks `(not (realized? (rest s)))` and `(not (realized? (nthrest* s 50)))`.

The other two gaps I flagged earlier (`first` on scalar body, `doall` of `(lazy-seq [v])`) turned out to already work on 0.38 — verified by `phel eval`. So once #1994 and #1995 land, the Phel branch should go green.

In the meantime: this PR can ship the `:phel (phel.core/lazy-seq? x)` line (after #884 lands) and the test file will simply stay red on Phel until upstream fixes follow.

test/clojure/core_test/lazy_seq.cljc
```diff
@@ -0,0 +1,72 @@
...
+
+      ;; The realized value of a lazy seq just has to be a seq, not
+      ;; necessarily a cons, so we can return a whole vector.
+      (let [s (lazy-seq [1 2 3])]
+        (is (p/lazy-seq? s))
+        (is (not (realized? s)))
+        (is (= [1 2 3] (doall s)))
+        (is (realized? s)))
+
+      (let [s (lazy-infinite-range)]
```
jeaye, line 48: Lazy seqs have an interesting property that nested lazy seqs and up getting flattened. This happens here: https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/LazySeq.java#L90

So, if we're realizing a lazy seq and we get another lazy seq, we enter the `unwrap` loop where we keep realizing until we get something which isn't a lazy seq. This might be an interesting case to test, since it's one which caused issues in jank when I didn't do it right.

jeaye (changes requested):
Nice work, Dave!

dgr:
> * [phel-lang/phel-lang#1994](https://github.com/phel-lang/phel-lang/issues/1994) — `next` on a lazy-seq returns a `LazySeq` instead of a realized seq → breaks `(not (p/lazy-seq? (next s)))`.

Let me think about this one. That might be OK and maybe I'm over-testing with respect to the JVM. The JVM definitely doesn't return a `lazy-seq`, but that's not necessarily mandated. But `next` should operate as `(seq (rest coll))`.

> * [phel-lang/phel-lang#1995](https://github.com/phel-lang/phel-lang/issues/1995) — `LazySeq::cdr` eagerly realizes the tail head → breaks `(not (realized? (rest s)))` and `(not (realized? (nthrest* s 50)))`.

`rest` should act basically like traditional Lisp `cdr`. Just return whatever is in the second half of the cell without evaluating it.

> The other two gaps I flagged earlier (`first` on scalar body, `doall` of `(lazy-seq [v])`) turned out to already work on 0.38 — verified by `phel eval`. So once #1994 and #1995 land, the Phel branch should go green.

Great, thanks!

> In the meantime: this PR can ship the `:phel (phel.core/lazy-seq? x)` line (after #884 lands) and the test file will simply stay red on Phel until upstream fixes follow.

Perfect. I'll add that.

test/clojure/core_test/lazy_seq.cljc
dgr: OK, let me think about that a bit. Definitely sounds like a good case.

dgr: Alright, new test added to validate case of nested lazy-seqs.

test/clojure/core_test/lazy_seq.cljc
```diff
@@ -0,0 +1,93 @@
...
+        (is (realized? (rest s))) ; `next` forces realization of next element
+        (is (p/lazy-seq? (nthrest* s 10)))
+        (is (not (realized? (nthrest* s 10))))))  ; but further elements not realized
+
+    (testing "oddball cases"
+      ;; You can have lazy seqs wrapping lazy seqs. In this case, when
+      ;; the lazy seq is realized, Clojure will keep realizing lazy
+      ;; seqs until it bottoms out with something that is not a lazy
+      ;; seq
+      (let [s (lazy-seq (lazy-seq (lazy-seq (cons 1 '(2 3)))))]
```
jeaye: These look great!

Am I being paranoid, or should we also wrap `'(2 3)` in a couple `lazy-seq` calls so we can test them being in the middle of things? I think jank ran into some issues with cons followed by nested lazy seqs, so I think it's worthwhile.

dgr: Sure, let's go crazy. ;-)

jeaye (approved):
Looks great to me, at this point. We're just blocked on the CI Pheliures. 🙃 

If Phel needs some time to fix things, we can just disable these tests with Phel for now. We cannot have a failing CI on main and we cannot merge failing CI for a PR.

dgr:
> Looks great to me, at this point. We're just blocked on the CI Pheliures. 🙃
> 
> If Phel needs some time to fix things, we can just disable these tests with Phel for now. We cannot have a failing CI on main and we cannot merge failing CI for a PR.

If we cannot merge with CI failing, I suggest that we strip CI down to just the basic dialects (e.g., JVM, bb, cljs, clr). If we don't do this, then we're going to move forward only as fast as the slowest dialect can update itself. I'd like to avoid that.

jeaye:
> > Looks great to me, at this point. We're just blocked on the CI Pheliures. 🙃
> > If Phel needs some time to fix things, we can just disable these tests with Phel for now. We cannot have a failing CI on main and we cannot merge failing CI for a PR.
> 
> If we cannot merge with CI failing, I suggest that we strip CI down to just the basic dialects (e.g., JVM, bb, cljs, clr). If we don't do this, then we're going to move forward only as fast as the slowest dialect can update itself. I'd like to avoid that.

I have no issue with disabling tests for certain dialects, with a TODO in the source. Perhaps we can just standardize on a format for those TODOs. We should still be able to move forward as fast as we like, tagging dialects which need to catch up as needed.

Maybe it's time to formalize a macro for this, such as `(p/skip-for-dialects #{:phel :jank} (deftest ...))`.

Thoughts?

jeaye:
The goal of formalizing a macro would be that we could use that macro to generate output for the parity of each dialect. We could then use this for various aspects throughout each test suite. This will move us one step closer to having a generated report from each CI run for which dialects can do what.

dgr:
Yea, we need something like that. We need some way to decouple work as much as possible. Further, we need a way to decouple knowledge of all dialects for test writers. We should not expect a test writer to be knowledgeable about the nuances of all dialects other than JVM and CLJS (IMO). We automated some of that with the `when-var-exists` macro to just skip tests for dialects that don't have a given var. But we need to do the same with failing tests where the var exists but the dialect implements things differently. What we don't want to have is a dialect maintainer goes on vacation and all merging stops with the test suite.

dgr:
> * [phel-lang/phel-lang#1994](https://github.com/phel-lang/phel-lang/issues/1994) — `next` on a lazy-seq returns a `LazySeq` instead of a realized seq → breaks `(not (p/lazy-seq? (next s)))`.
> * [](https://github.com/phel-lang/phel-lang/issues/1995)

@Chemaclass , I removed this test, so your call on whether you want to adopt the identical behavior of Clojure JVM or not. In this case, I don't think it's required that you behave identically.

Chemaclass:
Status from Phel side: working on [phel-lang/phel-lang#1994](https://github.com/phel-lang/phel-lang/issues/1994) and https://github.com/phel-lang/phel-lang/issues/1995. 
                                                                                                                                                        
Agree with you both: one dialect shouldn't block the suite (vacation scenario is exactly the failure mode). Per-test skip keeps parity pressure; stripping CI to core dialects drops it.                                                                                                               
                                                                                                                                                        
Proposal:                                                                                                                                             
1. Skip failing `lazy-seq` assertions on `:phel` with TODO or similar
2. Merge once green
3. I'll open a follow-up bumping Phel + removing skip + restoring the `(next s)` assertion once the Phel PR lands

jeaye:
Does anyone here have the interest and capacity to develop out the per-dialect parity report, based on the proposed macro usage (and `when-var-exists`)? This is something which we can generate and then continuously publish to GH pages in this repo so we have a dashboard of which dialects can do what.

@E-A-Griffin for vis, in case you want it.

E-A-Griffin:
> Does anyone here have the interest and capacity to develop out the per-dialect parity report, based on the proposed macro usage (and `when-var-exists`)? This is something which we can generate and then continuously publish to GH pages in this repo so we have a dashboard of which dialects can do what.
> 
> @E-A-Griffin for vis, in case you want it.

@jeaye what all would that entail?

jeaye:
> > Does anyone here have the interest and capacity to develop out the per-dialect parity report, based on the proposed macro usage (and `when-var-exists`)? This is something which we can generate and then continuously publish to GH pages in this repo so we have a dashboard of which dialects can do what.
> > @E-A-Griffin for vis, in case you want it.
> 
> @jeaye what all would that entail?

The simplest starting point is this:

0. Start tracking state in our test runner for which tests are run vs skipped
1. Add a `(p/skip-for-dialects #{:phel :jank} (deftest ...))` to track that state
2. Also track skips from `when-var-exists` being false
3. Generate a report for each dialect, in CI, and export that as a CI artifact

This would already be helpful to have and could be merged. From there, we can:

1. Use a job which runs after all dialects to combine the artifacts into a single artifact
2. Turn this artifact into a helpful table and write it to `gh-pages` so we can view it easily from a browser
3. Hook this into the suite at a per-case level, rather than just per-test, so we can cover cases like "+ works for CLJS, but ratios aren't supported"
  a. This can end up including another macro, if needed, to capture why a particular case is skipped, which ends up int he report

As an example of what I mean, Gloat has pretty much exactly this, in its test report here: https://gloathub.org/test-report/ We would just want to apply the same methodology to all dialects and ideally get some tables so that we can more easily diff parity across the board.

E-A-Griffin:
> > > Does anyone here have the interest and capacity to develop out the per-dialect parity report, based on the proposed macro usage (and `when-var-exists`)? This is something which we can generate and then continuously publish to GH pages in this repo so we have a dashboard of which dialects can do what.
> > > @E-A-Griffin for vis, in case you want it.
> > 
> > @jeaye what all would that entail?
> 
> The simplest starting point is this:
> 
> 0. Start tracking state in our test runner for which tests are run vs skipped
> 1. Add a `(p/skip-for-dialects #{:phel :jank} (deftest ...))` to track that state
> 2. Also track skips from `when-var-exists` being false
> 3. Generate a report for each dialect, in CI, and export that as a CI artifact
> 
> This would already be helpful to have and could be merged. From there, we can:
> 
> 1. Use a job which runs after all dialects to combine the artifacts into a single artifact
> 2. Turn this artifact into a helpful table and write it to `gh-pages` so we can view it easily from a browser
> 3. Hook this into the suite at a per-case level, rather than just per-test, so we can cover cases like "+ works for CLJS, but ratios aren't supported"
>   a. This can end up including another macro, if needed, to capture why a particular case is skipped, which ends up int he report
> 
> As an example of what I mean, Gloat has pretty much exactly this, in its test report here: https://gloathub.org/test-report/ We would just want to apply the same methodology to all dialects and ideally get some tables so that we can more easily diff parity across the board.

Yeah this would be good to have I can't immediately pick it up so we should probably make an issue on GH for it and let someone claim it when they have capacity (could be me at some later point)

test/clojure/core_test/portability.cljc
```diff
@@ -32,6 +32,14 @@
       :phel #(phel.async/delay (/ % 1000)))
    ms))
 
+(defn lazy-seq?
+  "Is `x` a lazy seq. This requires a dialect-specific test of some
+  sort, typically for whether `x` conforms to a specific object type."
+  [x]
+  #?(:cljs (instance? LazySeq x)
+     :phel (phel.core/lazy-seq? x)
+     :default (instance? clojure.lang.LazySeq x)))
```
chrisrink10, line 41:
```
:lpy (instance? basilisp.lang.seq/LazySeq x)
```

For Basilisp.

PR #888, opened 2026-05-20
Simplify `p/thrown?` code for various dialects

dgr:
Reworks the dialect-specific definitions of `p/thrown?` in `portability.cljc`. Removes lots of fine-grained conditional reader tags and moves to a one-definition-per-dialect model.

### DIFF
--- a/test/clojure/core_test/portability.cljc
+++ b/test/clojure/core_test/portability.cljc
@@ -112,7 +112,24 @@
                             :actual actual#})
               actual#)))))
 
-   :default  ; Clojure JVM, CLR, Babashka, Phel
+   :cljr
+   (defmethod t/assert-expr 'p/thrown?
+     [msg form]
+     (let [body (rest form)]
+       `(try
+          (let [result# (do ~@body)]
+            (t/do-report {:type :fail
+                          :message ~msg
+                          :expected '~form
+                          :actual result#}))
+          (catch ~'Exception e#
+            (t/do-report {:type :pass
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/portability.cljc
```diff
@@ -50,65 +50,103 @@
...
-                    (report-success# (success-opts# (~'.what e#))))))))))
+                          :actual result#}))
+          (catch ~'Exception e#
+            (t/do-report {:type :pass
+                          :message ~msg
+                          :expected '~form
+                          :actual e#})
+            e#))))
+
+   :default  ; Clojure JVM, Babashka, Phel
```
E-A-Griffin, line 110: nit: I think this comment is apt to not be maintained as dialects are added and is probably better off just removed

E-A-Griffin (approved):
Left a small nit but mostly looks good to me

test/clojure/core_test/portability.cljc
dgr: You might be right about maintenance. I felt like I wanted to make sure that the major dialects we're dealing with are accounted for.

jeaye (approved):
LGTM. Are you pushing a change for Emma's nit, or leaving it?

dgr:
> LGTM. Are you pushing a change for Emma's nit, or leaving it?

I was going to leave it.

jeaye:
Thanks, Dave!

PR #889, opened 2026-05-20
Add tests for `map`

### LINKED ISSUE #351: `clojure.core/map`

dgr:
Closes #351

### DIFF
--- a/test/clojure/core_test/map.cljc
+++ b/test/clojure/core_test/map.cljc
@@ -27,6 +27,12 @@
         (is (and (p/lazy-seq? try-nil) (empty? try-nil)))
         (is (and (p/lazy-seq? try-empty-list) (empty? try-empty-list))))
 
+      ;; infinite lazy seq
+      (let [s (map + (range))]
+        (is (p/lazy-seq? s))
+        (is (not (realized? s)))
+        (is (= [0 1 2 3 4] (take 5 s))))
+
       ;; Works on anything that can be turned into a seq, but order
       ;; isn't guaranteed for things like maps or sets
       (is (= [[:k :v]] (map identity {:k :v})))
@@ -45,11 +51,22 @@
       ;; one that is
       (is (= [0 2 4 6] (map + (range 5) (range 4))))
       (is (= [0 2 4 6] (map + (range 4) (range 5))))
+
+      ;; Infinite ranges
+      (let [s (map + (range) (range))]
+        (is (p/lazy-seq? s))
+        (is (not (realized? s)))
+        (is (= [0 2 4 6 8] (take 5 s))))
+      
       ;; if one of the seqs is empty, map yields no elements
       (is (empty? (map + (range 5) nil)))
+      (is (empty? (map + (range) nil)))
       (is (empty? (map + nil (range 5))))
+      (is (empty? (map + nil (range))))
       (is (empty? (map + (range 5) '())))
+      (is (empty? (map + (range) '())))
       (is (empty? (map + '() (range 5))))
+      (is (empty? (map + '() (range))))
       (is (empty? (map + nil nil)))
       (is (empty? (map + '() '()))))
     
@@ -66,14 +83,18 @@
       (is (= [0 3 6 9] (map + (range 5) (range 5) (range 4))))
       (is (= [0 3 6 9] (map + (range 5) (range 4) (range 5))))
       (is (= [0 3 6 9] (map + (range 4) (range 5) (range 5))))
+
+      ;; Infinite ranges
+      (is (= [0 3 6 9] (take 4 (map + (range) (range) (range)))))
+      
       ;; if any of the seqs is `nil`, result is empty
-      (is (empty? (map + (range 5) (range 5) nil)))
-      (is (empty? (map + (range 5) nil (range 5))))
-      (is (empty? (map + nil (range 5) (range 5))))
+      (is (empty? (map + (range 5) (range) nil)))
+      (is (empty? (map + (range 5) nil (range))))
+      (is (empty? (map + nil (range 5) (range))))
       ;; if any of the seqs is the empty list, result is empty
-      (is (empty? (map + (range 5) (range 5) '())))
-      (is (empty? (map + (range 5) '() (range 5))))
-      (is (empty? (map + '() (range 5) (range 5)))))
+      (is (empty? (map + (range 5) (range) '())))
+      (is (empty? (map + (range 5) '() (range))))
+      (is (empty? (map + '() (range 5) (range)))))
 
     (testing "arity 5+"
       ;; multiple seqs result in multiple args to the mapping function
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/map.cljc
```diff
@@ -0,0 +1,115 @@
...
+        ;; Does it transduce correctly?
+        (is (= [1 2 3 4 5] (transduce mfn conj (range 5))))
+        ;; Can it be composed?
+        (is (= [2 3 4 5 6] (transduce (comp mfn mfn) conj (range 5))))))
+    
+    (testing "arity 2"
+      (let [result (map inc (range 5))]
+        ;; `map` should return an unrealized lazy-seq
+        (is (p/lazy-seq? result))
+        (realized? result)
```
E-A-Griffin: missing an `is` here

test/clojure/core_test/map.cljc
```diff
@@ -0,0 +1,115 @@
...
+        (is (seq? result))
```
E-A-Griffin: what is the point of this if we check `lazy-seq?` before?

test/clojure/core_test/map.cljc
```diff
@@ -0,0 +1,115 @@
...
+        (is (= [1 2 3 4 5] result)))
+
+      ;; nil and empty list just result in empty seq
+      (let [try-nil (map identity nil)
+            try-empty-list (map identity '())]
+        (is (and (p/lazy-seq? try-nil) (empty? try-nil)))
+        (is (and (p/lazy-seq? try-empty-list) (empty? try-empty-list))))
```
E-A-Griffin, line 28: I think checking `lazy-seq?` here is redundant but won't block on it

test/clojure/core_test/map.cljc
```diff
@@ -0,0 +1,115 @@
...
+      ;; if any of the seqs is `nil`, result is empty
+      (is (empty? (map + (range 5) (range 5) nil)))
+      (is (empty? (map + (range 5) nil (range 5))))
+      (is (empty? (map + nil (range 5) (range 5))))
+      ;; if any of the seqs is the empty list, result is empty
+      (is (empty? (map + (range 5) (range 5) '())))
+      (is (empty? (map + (range 5) '() (range 5))))
+      (is (empty? (map + '() (range 5) (range 5)))))
+
+    (testing "arity 5+"
```
E-A-Griffin, line 99: I think we should have at least one test with an arity of 6 or greater in here since arity 5 is the minimal `&` arity

E-A-Griffin (commented):
Left some suggestions

test/clojure/core_test/map.cljc
dgr: Duh. Thank you.

test/clojure/core_test/map.cljc
dgr: Probably redundant. I'll remove.

test/clojure/core_test/map.cljc
dgr: This one I think is valid. I want to ensure that even with an empty or `nil` list, the return value of `map` is still a lazy seq. That is, no shortcuts being taken where the implementation just returns the empty list or `nil`.

test/clojure/core_test/map.cljc
dgr: OK

test/clojure/core_test/map.cljc
dgr: Fixed.

test/clojure/core_test/map.cljc
```diff
@@ -0,0 +1,115 @@
...
+      ;; Returns transducer
+      (let [mfn (map inc)]
+        ;; Is it a function?
+        (is (fn? mfn))
+        ;; Does it transduce correctly?
+        (is (= [1 2 3 4 5] (transduce mfn conj (range 5))))
+        ;; Can it be composed?
+        (is (= [2 3 4 5 6] (transduce (comp mfn mfn) conj (range 5))))))
+    
+    (testing "arity 2"
```
jeaye, line 17: I believe we're missing a case for `map` with an infinite seq, which returns another infinite seq. For example:

```clojure
(map + (range))
```

This probably makes sense to do with other arities, too. For example:

```clojure
(map + (range) (range))
```

dgr:
I also added a test case where some of the collections are infinite `range`s, to validate that it doesn't try to realize them  fully.

test/clojure/core_test/map.cljc
```diff
@@ -0,0 +1,115 @@
...
+      ;; if any of the seqs is the empty list, result is empty
+      (is (empty? (map + (range 5) (range 5) (range 5) '())))
+      (is (empty? (map + (range 5) (range 5) '() (range 5))))
+      (is (empty? (map + (range 5) '() (range 5) (range 5))))
+      (is (empty? (map + '() (range 5) (range 5) (range 5)))))
+
+    (testing "negative cases"
+      ;; Note: must realize resulting lazy-seq with `seq` to force the
+      ;; exception
+      (are [x] (p/thrown? (seq (map identity x)))
```
jeaye: I don't think `seq` is the correct fn to use to force here, conceptually. I think we want `doall`.

jeaye (changes requested):
Great work, Dave!

test/clojure/core_test/map.cljc
dgr: It doesn’t matter which function you use to force it. `Seq` works just as well as `doall`. But if you want me to use `doall`, I can.

test/clojure/core_test/map.cljc
dgr: Agreed. Will add it.

dgr: Done. Added lots of infinite seas throughout.

dgr:
Added lots of infinite ranges. Did not convert to `doall` from `seq`. If you want me to do that, let me know.

test/clojure/core_test/map.cljc
E-A-Griffin: I do think if this is preferred we should make it an explicit recommendation in the writing tests doc markdown file so CoPilot can catch it and also so new test writers know it's the preferred function for realization of lazy seqs

jeaye: I know that `seq`, for some seqs, does nothing and just returns the same seq. Whereas `doall` is explicitly used to force the realization of a full seq for the purposes of effects.

`doall` just expands to `dorun`, which does a loop through the whole seq.

```clojure
(defn dorun
  "When lazy sequences are produced via functions that have side
  effects, any effects other than those needed to produce the first
  element in the seq do not occur until the seq is consumed. dorun can
  be used to force any effects. Walks through the successive nexts of
  the seq, does not retain the head and returns nil."
  {:added "1.0"
   :static true}
  ([coll]
   (when-let [s (seq coll)]
     (recur (next s))))
  ([n coll]
   (when (and (seq coll) (pos? n))
     (recur (dec n) (next coll)))))
```

Whereas just calling `seq` is not going to realize the full thing. So, if there's an error at the end of a lazy seq, calling `seq` on it may not trigger a throw, but calling `doall` will. That's the key distinction.

Thoughts?

dgr: Again, I don't know that it matters in this case since we know that the lazy seq is just a single element which when realized with throw because the just-realized object isn't a seq. In other words, yea, sure, `doall` is what you use when you want to realize a whole seq, but in this case, realizing the first element is enough because the first element is bogus. And we know that because we're forcing it to be bogus in the test. But fine, I'll use `doall`.

dgr: OK, done.

jeaye (approved):
Phew! Excellent work, Dave.

PR #890, opened 2026-05-21
Add tests for `concat`

### LINKED ISSUE #196: `concat`

dgr:
Closes #196

### DIFF
--- a/test/clojure/core_test/concat.cljc
+++ b/test/clojure/core_test/concat.cljc
@@ -42,7 +42,11 @@
         (is (p/lazy-seq? s))
         (is (not (realized? s)))
         ;; Now, take the first 16 items from the concatenation
-        (is (= [0 1 2 3 4 5 6 7 8 9 10 11 12 0 1 2] (take 16 s)))))
+        (is (= [0 1 2 3 4 5 6 7 8 9 10 11 12 0 1 2] (take 16 s))))
+
+      ;; Go crazy with ranges, empty list, and nil
+      (is (= [0 1 2 3 4 5 0 1 2 3 4 5 6 7 8]
+             (take 15 (concat (range 3) nil (range 3 6) '() (range))))))
 
     (testing "empty list and nil"
       (is (= [1 2 3] (concat [1 2 3] '())))
--- a/test/clojure/core_test/concat.cljc
+++ b/test/clojure/core_test/concat.cljc
@@ -1,0 +1,51 @@
+(ns clojure.core-test.concat
+  (:require [clojure.test :as t :refer [are deftest is testing]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists] :as p]))
+
+(when-var-exists concat
+  (deftest test-concat
+    (testing "arity 0"
+      ;; returns an unrealized, empty lazy seq
+      (let [s (concat)]
+        (is (p/lazy-seq? s))
+        (is (not (realized? s)))
+        (is (empty? s))))
+
+    (testing "arity 1"
+      ;; returns an unrealized lazy seq of just the elements in the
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/concat.cljc
```diff
@@ -0,0 +1,51 @@
...
+        (is (not (realized? s)))
+        (is (= [1 2 3 4 5 6 7 8 9 10] s)))
+
+      ;; Infinite range as second argument.
+      (is (= [1 2 3 0 1 2] (take 6 (concat [1 2 3] (range)))))
+      ;; Infinite range as first argument, so we'll never take
+      ;; elements from the second argument.
+      (is (= [0 1 2 3 4] (take 5 (concat (range) [:a :b :c])))))
+
+    (testing "arity 3+"
```
jeaye, line 39: I think `(concat s1 nil s2 nil s3)` is an interesting potential case, since some of the interim seqs are nil.

jeaye (changes requested):

E-A-Griffin (commented):
Let's add tests for other collection types (e.g. (sorted/unsorted) sets, (sorted/unsorted) maps, mix of sequential and associative collections, maybe even ranges)

E-A-Griffin (changes requested):
Let's add tests for other collection types (e.g. (sorted/unsorted) sets, (sorted/unsorted) maps, mix of sequential and associative collections, maybe even ranges)

test/clojure/core_test/concat.cljc
E-A-Griffin: ^

dgr: Yes, I like that. I'll add.

dgr:
> Let's add tests for other collection types (e.g. (sorted/unsorted) sets, (sorted/unsorted) maps, mix of sequential and associative collections, maybe even ranges)

So, I debated about this for quite a while. In general, I think those tests are perhaps overused. Everything gets converted to a `seq` by these collection functions. The point isn't to test whether collection of type X can be converted into a `seq`, but rather whether `concat` runs correctly with the seqs it is given.

Further, when you test sorted/unsorted maps and sets and such, you end up having to involve more functions in the tests themselves (e.g., have to convert everything to sets to compare them because order is not guaranteed for sets and maps). This means more demands on the system under test in order to run the test. I've generally been trying to keep a low set of demands for running tests, trying to rely on as little as possible.

So, in this case, I decided that the juice wasn't worth the squeeze. @jeaye , thoughts here?

test/clojure/core_test/concat.cljc
dgr: Done.

jeaye:
> > Let's add tests for other collection types (e.g. (sorted/unsorted) sets, (sorted/unsorted) maps, mix of sequential and associative collections, maybe even ranges)
> 
> So, I debated about this for quite a while. In general, I think those tests are perhaps overused. Everything gets converted to a `seq` by these collection functions. The point isn't to test whether collection of type X can be converted into a `seq`, but rather whether `concat` runs correctly with the seqs it is given.
> 
> Further, when you test sorted/unsorted maps and sets and such, you end up having to involve more functions in the tests themselves (e.g., have to convert everything to sets to compare them because order is not guaranteed for sets and maps). This means more demands on the system under test in order to run the test. I've generally been trying to keep a low set of demands for running tests, trying to rely on as little as possible.
> 
> So, in this case, I decided that the juice wasn't worth the squeeze. @jeaye , thoughts here?

Yes, I agree. Throwing different collections, all of which will be turned into seqs, will not help exercise the function.

dgr:
I believe this is ready for merging.

jeaye:
Thank you, Dave!

PR #893, opened 2026-05-22
Add tests for `partition`

### LINKED ISSUE #389: `clojure.core/partition`

dgr:
Closes #389

### DIFF
--- a/test/clojure/core_test/partition.cljc
+++ b/test/clojure/core_test/partition.cljc
@@ -40,7 +40,7 @@
       (is (= '() (partition 2 1 '())))
       (is (= '() (partition 2 1 nil))))
 
-    (testing "arity 4 - (partition n step pad coll"
+    (testing "arity 4 - (partition n step pad coll)"
       ;; Use padding for the last element
       (is (= '((0 1 2) (1 2 3) (2 3 4) (3 4 :a)) (partition 3 1 [:a :b :c] (range 5))))
       (is (= '((0 1 2) (3 4 5) (6 7 8) (9 :a :b))  (partition 3 3 [:a :b :c] (range 10))))
--- a/test/clojure/core_test/partition.cljc
+++ b/test/clojure/core_test/partition.cljc
@@ -19,6 +19,16 @@
         (is (not (realized? s)))
         (is (= '((0 1) (2 3) (4 5) (6 7) (8 9)) s)))
 
+      ;; The partition size, `n` should be able to be a number of all sorts.
+      (is (= '((0 1) (2 3) (4 5) (6 7) (8 9)) (partition 2N (range 10))))
+
+      ;; These fail in JVM and some other dialects but should probably
+      ;; pass. Interestingly, since CLJS uses doubles as its only
+      ;; numeric type, they pass in CLJS.
+      ;; TODO: reenable these for JVM and other dialects when they pass
+      #?@(:cljs [(is (= '((0 1) (2 3) (4 5) (6 7) (8 9)) (partition 2.0 (range 10))))
+                 (is (= '((0 1) (2 3) (4 5) (6 7) (8 9)) (partition 2.0M (range 10))))])
+
       ;; Infinite range
       (is (= '((0 1) (2 3) (4 5) (6 7)) (take 4 (partition 2 (range)))))
 
@@ -32,6 +42,11 @@
       ;; Note that we tried testing non-positive steps, but that led
       ;; to an infinite loop within `partition`.
       (is (= '((0 1) (1 2) (2 3) (3 4) (4 5)) (partition 2 1 (range 6))))
+
+      ;; Test steps of different number types
+      (is (= '((0 1) (1 2) (2 3) (3 4) (4 5)) (partition 2 1N (range 6))))
+      (is (= '((0 1) (1 2) (2 3) (3 4) (4 5)) (partition 2 1.0 (range 6))))
+      (is (= '((0 1) (1 2) (2 3) (3 4) (4 5)) (partition 2 1.0M (range 6))))
 
       ;; Try an infinite range
       (is (= '((0 1) (1 2) (2 3) (3 4) (4 5)) (take 5 (partition 2 1 (range)))))
--- a/test/clojure/core_test/partition.cljc
+++ b/test/clojure/core_test/partition.cljc
@@ -42,8 +42,17 @@
 
     (testing "arity 4 - (partition n step pad coll"
       ;; Use padding for the last element
-      (is (= '((0 1 2) (1 2 3) (2 3 4) (3 4 :a)) (partition 3 1 [:a :a :a] (range 5))))
-      (is (= '((0 1 2) (3 4 5) (6 7 8) (9 :a :a))  (partition 3 3 [:a :a :a] (range 10))))
+      (is (= '((0 1 2) (1 2 3) (2 3 4) (3 4 :a)) (partition 3 1 [:a :b :c] (range 5))))
+      (is (= '((0 1 2) (3 4 5) (6 7 8) (9 :a :b))  (partition 3 3 [:a :b :c] (range 10))))
+
+      ;; If the padding collection is short, then the last partition is short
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/partition.cljc
```diff
@@ -0,0 +1,50 @@
...
+
+    ;; Docstring:
+    ;; [n coll] [n step coll] [n step pad coll]
+    ;; Returns a lazy sequence of lists of n items each, at offsets step
+    ;; apart. If step is not supplied, defaults to n, i.e. the partitions
+    ;; do not overlap. If a pad collection is supplied, use its elements as
+    ;; necessary to complete last partition upto n items. In case there are
+    ;; not enough padding elements, return a partition with less than n items.
+
+    (testing "arity 2 - (partition n coll)"
```
jeaye, line 16: Potential tests: `n` is zero or negative

Potential tests: `step` is zero or negative

Potential tests: `pad` is zero or negative

test/clojure/core_test/partition.cljc
```diff
@@ -0,0 +1,50 @@
...
+      (is (= '((0 1) (1 2) (2 3) (3 4) (4 5)) (take 5 (partition 2 1 (range)))))
+
+      ;; empty list and nil
+      (is (= '() (partition 2 1 '())))
+      (is (= '() (partition 2 1 nil))))
+
+    (testing "arity 4 - (partition n step pad coll"
+      ;; Use padding for the last element
+      (is (= '((0 1 2) (1 2 3) (2 3 4) (3 4 :a)) (partition 3 1 [:a :a :a] (range 5))))
+      (is (= '((0 1 2) (3 4 5) (6 7 8) (9 :a :a))  (partition 3 3 [:a :a :a] (range 10))))
```
jeaye: Potential test: What if pad collection is too small?

Potential test: What if pad collection is empty?

test/clojure/core_test/partition.cljc
```diff
@@ -0,0 +1,50 @@
...
+      ;; Try an infinite range
+      (is (= '((0 1) (1 2) (2 3) (3 4) (4 5)) (take 5 (partition 2 1 (range)))))
+
+      ;; empty list and nil
+      (is (= '() (partition 2 1 '())))
+      (is (= '() (partition 2 1 nil))))
+
+    (testing "arity 4 - (partition n step pad coll"
+      ;; Use padding for the last element
+      (is (= '((0 1 2) (1 2 3) (2 3 4) (3 4 :a)) (partition 3 1 [:a :a :a] (range 5))))
```
jeaye: Let's use different values in the pad, to catch any issues around the order that they're used. Just `[:a :b :c]` should be fine.

test/clojure/core_test/partition.cljc
```diff
@@ -0,0 +1,50 @@
+(ns clojure.core-test.partition
+  (:require [clojure.test :as t :refer [are deftest is testing]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists] :as p]))
+
+(when-var-exists partition
+  (deftest test-partition
+
+    ;; Docstring:
+    ;; [n coll] [n step coll] [n step pad coll]
```
jeaye, line 9: What do we think about testing `n` and `step` being something more interesting than a `Long`? Such as a big integer? Still fits within the valid input assumption, I think, so worthwhile.

jeaye (changes requested):
Nice work!

test/clojure/core_test/partition.cljc
dgr: I checked `n` and `step` being zero or negative and those result in infinite loops, so I pulled those tests. Pad is a collection of items, not a number. Items from the pad collection are added to the last group. This is tested. That said, I added tests for when the padding is a short collection (fewer items in the padding collection than required to pad out the final partition) as well as empty and `nil` (just special cases of short padding, really).

test/clojure/core_test/partition.cljc
dgr: Yes, added.

test/clojure/core_test/partition.cljc
```diff
@@ -0,0 +1,50 @@
...
+      (is (= '((0 1) (1 2) (2 3) (3 4) (4 5)) (partition 2 1 (range 6))))
+
+      ;; Try an infinite range
+      (is (= '((0 1) (1 2) (2 3) (3 4) (4 5)) (take 5 (partition 2 1 (range)))))
+
+      ;; empty list and nil
+      (is (= '() (partition 2 1 '())))
+      (is (= '() (partition 2 1 nil))))
+
+    (testing "arity 4 - (partition n step pad coll"
```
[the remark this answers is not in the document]
dgr: Fixed.

test/clojure/core_test/partition.cljc
```diff
@@ -0,0 +1,50 @@
...
+(when-var-exists partition
+  (deftest test-partition
+
+    ;; Docstring:
+    ;; [n coll] [n step coll] [n step pad coll]
+    ;; Returns a lazy sequence of lists of n items each, at offsets step
+    ;; apart. If step is not supplied, defaults to n, i.e. the partitions
+    ;; do not overlap. If a pad collection is supplied, use its elements as
+    ;; necessary to complete last partition upto n items. In case there are
+    ;; not enough padding elements, return a partition with less than n items.
```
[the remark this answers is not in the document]
dgr: Not going to resolve.

test/clojure/core_test/partition.cljc
dgr: Sure, I can try those.

dgr: Done.

dgr:
I believe this is ready to merge.

jeaye:
I'm not convinced we want the TODO for the floating point value support, since I gather that the fn is intended to only accept integral values. CLJS just happens to support it since there's only one number type for JS.

I think we should remove the floating point case, and the TODO. Keeping the big integer case is good, though, since it's an integral type.

E-A-Griffin:
>I'm not convinced we want the TODO for the floating point value support, since I gather that the fn is intended to only accept integral values. CLJS just happens to support it since there's only one number type for JS.

> I think we should remove the floating point case, and the TODO. Keeping the big integer case is good, though, since it's an integral type.

I think the test is good to keep but the TODO should be dropped as I don't think non CLJS are obligated to do any particular behavior with floats, that being said, I think CLJS' behavior is interesting and worth documenting here (could see this causing a real bug in the wild for someone)

E-A-Griffin (commented):
I think we should add a test for `(partition 0 ...)` akin to what's shown here: https://clojuredocs.org/clojure.core/partition#example-57aa75fae4b0bafd3e2a04d3

E-A-Griffin (commented):
Mostly looks good, left some optional suggestions

dgr:
> I think we should add a test for `(partition 0 ...)` akin to what's shown here: https://clojuredocs.org/clojure.core/partition#example-57aa75fae4b0bafd3e2a04d3

Sorry for the lag. Been busy with other things.

I originally tried some tests like that with zero and negative values for `n` and `step`, but they hang or repeat infinite values, which made me rethink them. Yes, the example at clojuredocs.org uses `take` to get the first few of an infinite stream of empty lists. But is that really useful? Is that really within the contract that `partition` is making? The assumption seems to be that `n` and `step` are positive integers, not zero or negative.

jeaye:
I think the only thing blocking this one is the TODO and CLJS tests for floating point steps. This is not the intended use of the fn and it's only because JS has one number type that it works on CLJS. I don't think there is anything TODO about supporting floating point steps for other dialects.

dgr:
> I think the only thing blocking this one is the TODO and CLJS tests for floating point steps. This is not the intended use of the fn and it's only because JS has one number type that it works on CLJS. I don't think there is anything TODO about supporting floating point steps for other dialects.

Done

jeaye:
Thanks, Dave!

PR #896, opened 2026-05-24
initial tests for clojure.core/merge-with

### LINKED ISSUE #361: `merge-with`

perrygeo:
Testing `clojure.core/merge-with`

## Manually tested with
- [x] JVM
- [x] Clojurescript
- [x] Clojure CLR
- [x] Babashka
- Basalisp (fails, see comment on nil handling)
- PHEL (I don't have access to PHP tooling at the moment)

## Dialect Differences

* Clojurescript returns a `nil` in the zero-arg case. Clojure etc throw an exception.
* Clojurescript: numeric keys have different equality semantics than the JVM.
* Basalisp returns `nil` on `(merge-with second-arg nil {})`, rest return `{}`

---

Anything else I can do to improve this? Let me know. This is a really fun way to get to know the clojure standard library!

Resolves #361

### DIFF
--- a/test/clojure/core_test/merge_with.cljc
+++ b/test/clojure/core_test/merge_with.cljc
@@ -1,0 +1,79 @@
+(ns clojure.core-test.merge-with
+  (:require [clojure.test :as t :refer [deftest is testing]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists] :as p]))
+
+(defn second-arg [_a b] b)
+
+(when-var-exists
+ merge-with
+ (deftest test-merge-with
+   (testing "`merge-with`"
+
+     (testing "zero arg behavior"
+       ;; NOTE: clojurescript differs from the rest here, bug?
+       #?(:clj     (is (p/thrown? (merge-with)))
+          :cljs    (is (nil? (merge-with))) ; TODO how to suppress warning
+          :default (is (p/thrown? (merge-with)))))
+
+     (testing "`nil`ish behavior"
+       (is (nil? (merge-with second-arg nil)))
+       (is (nil? (merge-with second-arg nil nil nil)))
+       (is (= {} (merge-with second-arg {} nil)))
+       (is (= {} (merge-with second-arg nil {}))) ;; fails in basilisp => nil
+       (is (= {:a 1} (merge-with second-arg {:a 1} nil {})))
+       (is (= {:a 1} (merge-with second-arg {:a 1} {} nil))))
+
+     (testing "f is applied left-to-right across maps"
+       ;; str is non-commutative, so order reveals the fold direction
+       (is (= {:a "123"} (merge-with str {:a "1"} {:a "2"} {:a "3"}))))
+
+     (testing "falsey values still trigger f on collision"
+       (is (= {:a false} (merge-with second-arg {:a true} {:a false})))
+       (is (= {:a nil} (merge-with second-arg {:a true} {:a nil}))))
+
+     (testing "when f returns nil, the key is kept with a nil value"
+       (is (= {:a nil} (merge-with (constantly nil) {:a 1} {:a 2}))))
+
+     (testing "f is only called on the values of duplicate keys"
+       (let [called-with (atom [])
+             tracking-sum (fn [a b]
+                            (swap! called-with conj [a b])
+                            (+ a b))]
+         ;; f is never called, no conflicts
+         (merge-with tracking-sum {:a 1 :b 10} {:c 100})
+         (is (= [] @called-with))
+         ;; f only called for values of :b
+         (merge-with tracking-sum {:a 1 :b 10} {:c 100 :b 20})
+         (is (= [[10 20]] @called-with))))
+
+     (testing "non-map types"
+       #?(:clj    (is (thrown-with-msg? java.lang.ClassCastException
+                                        #".*cannot be cast to .*"
+                                        (merge-with second-arg {:a 1} [:a 2])))
+          :cljs    (is (p/thrown? (merge-with second-arg {:a 1} [:a 2])))
+          :default (is (p/thrown? (merge-with second-arg {:a 1} [:a 2]))))
+       #?(:clj    (is (thrown-with-msg? java.lang.IllegalArgumentException
+                                        #"^Don't know how to create ISeq from.*"
+                                        (merge-with second-arg {:a 1} 1)))
+          :cljs    (is (p/thrown? (merge-with second-arg {:a 1} 1)))
+          :default (is (p/thrown? (merge-with second-arg {:a 1} 1)))))
+
+     (testing "duplicate mappings are handled according to f"
+       (is (= {:a 2} (merge-with second-arg {:a 1} {:a 2})))
+       (is (= {:a 3} (merge-with + {:a 1} {:a 2})))
+       (let [val 2]
+         (is (= {:a 4} (merge-with + {:a val} {:a val})))) ; not fooled by identity
+       (is (= {:a 3 :b 1} (merge-with + {:a 1} {:b 1} nil {:a 2})))
+       (is (= {:a [1 2] :b 3} (merge-with (fn [a b] (if (vector? a)
+                                                      (conj a b)
+                                                      [a b]))
+                                          {:a 1} {:a 2} {:b 3}))))
+
+     (testing "nested maps are merged with `into`"
+       (is (= {:ceo {:salary 1000000, :name "Alice"},
+               :cto {:salary 500000, :name "Brenda"}}
+              (merge-with into
+                          {:ceo {:salary 1000000}}
+                          {:cto {:salary  500000}}
+                          {:ceo {:name "Alice"}}
+                          {:cto {:name "Brenda"}})))))))
--- a/test/clojure/core_test/merge_with.cljc
+++ b/test/clojure/core_test/merge_with.cljc
@@ -10,9 +10,8 @@
    (testing "`merge-with`"
 
      (testing "zero arg behavior"
-       #?(:clj     (is (p/thrown? (merge-with)))
-          :cljs    (is (nil? (merge-with))) ; TODO how to suppress warning
-          :default (is (p/thrown? (merge-with)))))
+       #?(:cljs    (is (nil? (merge-with)))
+          :clj     (is (p/thrown? (merge-with)))))
 
      (testing "`nil`ish behavior"
        (is (nil? (merge-with second-arg nil)))
@@ -46,15 +45,15 @@
          (is (= [[10 20]] @called-with))))
 
      (testing "non-map types"
-       #?(:clj    (is (thrown-with-msg? java.lang.ClassCastException
+       #?(:cljs    (is (p/thrown? (merge-with second-arg {:a 1} [:a 2])))
+          :clj    (is (thrown-with-msg? java.lang.ClassCastException
                                         #".*cannot be cast to .*"
                                         (merge-with second-arg {:a 1} [:a 2])))
-          :cljs    (is (p/thrown? (merge-with second-arg {:a 1} [:a 2])))
           :default (is (p/thrown? (merge-with second-arg {:a 1} [:a 2]))))
-       #?(:clj    (is (thrown-with-msg? java.lang.IllegalArgumentException
+       #?(:cljs    (is (p/thrown? (merge-with second-arg {:a 1} 1)))
+          :clj    (is (thrown-with-msg? java.lang.IllegalArgumentException
                                         #"^Don't know how to create ISeq from.*"
                                         (merge-with second-arg {:a 1} 1)))
-          :cljs    (is (p/thrown? (merge-with second-arg {:a 1} 1)))
           :default (is (p/thrown? (merge-with second-arg {:a 1} 1)))))
 
      (testing "duplicate mappings are handled according to f"
--- a/test/clojure/core_test/merge_with.cljc
+++ b/test/clojure/core_test/merge_with.cljc
@@ -10,7 +10,6 @@
    (testing "`merge-with`"
 
      (testing "zero arg behavior"
-       ;; NOTE: clojurescript differs from the rest here, bug?
        #?(:clj     (is (p/thrown? (merge-with)))
           :cljs    (is (nil? (merge-with))) ; TODO how to suppress warning
           :default (is (p/thrown? (merge-with)))))
@@ -19,7 +18,7 @@
        (is (nil? (merge-with second-arg nil)))
        (is (nil? (merge-with second-arg nil nil nil)))
        (is (= {} (merge-with second-arg {} nil)))
-       (is (= {} (merge-with second-arg nil {}))) ;; fails in basilisp => nil
+       (is (= {} (merge-with second-arg nil {})))
        (is (= {:a 1} (merge-with second-arg {:a 1} nil {})))
        (is (= {:a 1} (merge-with second-arg {:a 1} {} nil))))
 
--- a/test/clojure/core_test/merge_with.cljc
+++ b/test/clojure/core_test/merge_with.cljc
@@ -45,22 +45,12 @@
          (is (= [[10 20]] @called-with))))
 
      (testing "non-map types"
-       #?(:cljs    (is (p/thrown? (merge-with second-arg {:a 1} [:a 2])))
-          :clj     (is (thrown-with-msg? java.lang.ClassCastException
-                                         #".*cannot be cast to .*"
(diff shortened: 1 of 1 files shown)


E-A-Griffin:
> basilisp and phel status is unknown, are the core 4 langs sufficient?

They are yeah, @jeaye disabled them for the time being, that being said noting how they fail is always nice for their respective maintainers to be aware of (and potentially fixed)

E-A-Griffin (changes requested):
Could you use the following reader conditional order: https://github.com/E-A-Griffin/clojure-test-suite/blob/e00fe2b8ff87fcf02020664cc57f11bbf4448904/doc/writing-tests.md#reader-conditional-order? This PR isn't merged yet but @jeaye and I agreed on it prior so going to try to maintain it going forward

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,79 @@
...
+
+(when-var-exists
+ merge-with
+ (deftest test-merge-with
+   (testing "`merge-with`"
+
+     (testing "zero arg behavior"
+       ;; NOTE: clojurescript differs from the rest here, bug?
+       #?(:clj     (is (p/thrown? (merge-with)))
+          :cljs    (is (nil? (merge-with))) ; TODO how to suppress warning
```
E-A-Griffin: What is this TODO for?

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,79 @@
...
+
+(defn second-arg [_a b] b)
+
+(when-var-exists
+ merge-with
+ (deftest test-merge-with
+   (testing "`merge-with`"
+
+     (testing "zero arg behavior"
+       ;; NOTE: clojurescript differs from the rest here, bug?
```
E-A-Griffin: Let's move this comment to the PR rather than in the code, CLJS often differs from other dialects and we should have a dialect differences report coming at some point

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,79 @@
...
+       #?(:clj     (is (p/thrown? (merge-with)))
+          :cljs    (is (nil? (merge-with))) ; TODO how to suppress warning
+          :default (is (p/thrown? (merge-with)))))
```
E-A-Griffin: if `:clj` and `:default` are the same, then we should just omit `:clj`

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,79 @@
...
+
+     (testing "`nil`ish behavior"
+       (is (nil? (merge-with second-arg nil)))
+       (is (nil? (merge-with second-arg nil nil nil)))
+       (is (= {} (merge-with second-arg {} nil)))
+       (is (= {} (merge-with second-arg nil {}))) ;; fails in basilisp => nil
+       (is (= {:a 1} (merge-with second-arg {:a 1} nil {})))
```
E-A-Griffin, line 21: not a hard requirement but if you know basilisp has different behavior, you might as well use a reader conditional here imo. Not block worthy though

E-A-Griffin: which would allow you to also remove the comment

E-A-Griffin:
I think `sorted-map` tests would also be nice

test/clojure/core_test/merge_with.cljc
perrygeo: When I run `bb test-cljs`, everything passes but the output contains this verbose warning:

```
...
Ran 223 tests containing 4957 assertions.
0 failures, 0 errors.
===============================================
[:test] Build completed. (324 files, 323 compiled, 1 warnings, 4.22s)

------ WARNING #1 - :fn-arity --------------------------------------------------
 File: /Users/matthew/projects/clojure-test-suite/test/clojure/core_test/merge_with.cljc:15:30
--------------------------------------------------------------------------------
  12 |      (testing "zero arg behavior"
  13 |        ;; NOTE: clojurescript differs from the rest here, bug?
  14 |        #?(:clj     (is (p/thrown? (merge-with)))
  15 |           :cljs    (is (nil? (merge-with))) ; TODO how to suppress warning
------------------------------------^-------------------------------------------
 Wrong number of args (0) passed to cljs.core/merge-with
--------------------------------------------------------------------------------
  16 |           :default (is (p/thrown? (merge-with)))))
  17 | 
  18 |      (testing "`nil`ish behavior"
  19 |        (is (nil? (merge-with second-arg nil)))
--------------------------------------------------------------------------------
```

Since we're explicitly testing the zero-args case, is there a way to silence this?

test/clojure/core_test/merge_with.cljc
perrygeo: I've removed the comment and moved it to the dialect differences section.

perrygeo:
Thanks for the review @E-A-Griffin ! The reader conditional ordering actually matters, TIL. I added tests for `sorted-map` and cleaned up a bit. 

The cljs zero-arg case is the only real question mark. I've noted the difference in the top PR description. The warning still bothers me though, hoping to find a way to suppress that.

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,85 @@
...
+     (testing "non-map types"
+       #?(:cljs    (is (p/thrown? (merge-with second-arg {:a 1} [:a 2])))
+          :clj     (is (thrown-with-msg? java.lang.ClassCastException
+                                         #".*cannot be cast to .*"
+                                         (merge-with second-arg {:a 1} [:a 2])))
+          :default (is (p/thrown? (merge-with second-arg {:a 1} [:a 2]))))
+       #?(:cljs    (is (p/thrown? (merge-with second-arg {:a 1} 1)))
+          :clj     (is (thrown-with-msg? java.lang.IllegalArgumentException
+                                         #"^Don't know how to create ISeq from.*"
+                                         (merge-with second-arg {:a 1} 1)))
```
jeaye: We don't want this `thrown-with-msg?` for two reasons:

1. Error messages can change and that should not break our tests
2. We do not want to duplicate tests just to provide a more specific exception type

What matters is that the function throws on specific inputs, not what it throws. So we just want to use `p/thrown?` here, which means there's no need for a reader conditional.

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,85 @@
...
+          :default (is (p/thrown? (merge-with second-arg {:a 1} 1)))))
+
+     (testing "duplicate mappings are handled according to f"
+       (is (= {:a 2} (merge-with second-arg {:a 1} {:a 2})))
+       (is (= {:a 3} (merge-with + {:a 1} {:a 2})))
+       (let [val 2]
+         (is (= {:a 4} (merge-with + {:a val} {:a val})))) ; not fooled by identity
```
jeaye: What do you mean "not fooled by identity" here? `merge-with` checks for collisions using the key, not the value, and keywords with the same value will already be identical. So basically every other test here is testing key identity.

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,85 @@
...
+                                          {:a 1} {:a 2} {:b 3}))))
+
+     (testing "sorted"
+       (let [result (merge-with +
+                                (sorted-map :c 3 :a 1 :b 2)
+                                (sorted-map :b 2 :c 3 :a 1))]
+         (is (= {:a 2 :b 4 :c 6} result))
+         (is (sorted? result))
+         (is (= [:a :b :c] (keys result)))))
+
```
jeaye, line 67: It would be good to have a test to see how keys are selected, when not identical, especially when metadata is involved. For example:

```clojure
user=> (meta (ffirst (merge-with + {^:a {} 1} {^:b {} 2})))
{:a true}
```

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,85 @@
+(ns clojure.core-test.merge-with
+  (:require [clojure.test :as t :refer [deftest is testing]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists] :as p]))
+
+(defn second-arg [_a b] b)
+
+(when-var-exists
+ merge-with
+ (deftest test-merge-with
```
jeaye, line 9: There's a lot more we can test here for keys which are equal. For example, these are some interesting cases.

```clojure
user=> (merge-with str {3 'long} {3N 'bigint})
{3 "longbigint"}

user=> (merge-with str {3 'long} {3.0 'double})
{3 long, 3.0 double}

user=> (merge-with str {3.0 'double} {3.0M 'bigdecimal})
{3.0 double, 3.0M bigdecimal}

user=> (merge-with str {(range 2) 'range} {'(0 1) 'list})
{(0 1) "rangelist"}
```

No doubt there are more.

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,85 @@
+(ns clojure.core-test.merge-with
+  (:require [clojure.test :as t :refer [deftest is testing]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists] :as p]))
+
+(defn second-arg [_a b] b)
```
jeaye, line 5: We can delete this fn and just use `second`.

jeaye (changes requested):
Thanks for the PR!

test/clojure/core_test/merge_with.cljc
perrygeo: yeah this test doesn't do what I was thinking, bad mental model. I'll remove it.

test/clojure/core_test/merge_with.cljc
perrygeo: `second` takes one arg and returns the 2nd element. I guess we could write `#(second [%1 %2])` but not sure if that's an improvement.

test/clojure/core_test/merge_with.cljc
perrygeo: Great idea. I'll add these verbatim and try to come up with some other interesting ones.

test/clojure/core_test/merge_with.cljc
perrygeo: In general I think of the possible exceptions thrown by a function as part of its public API. It's certainly something people write code against. Testing that any old exception is thrown feels a bit lacking. But I agree that *messages* aren't what we should match on. I wish there was a `thrown-with-exception?` that just asserted a type. But yeah, for now I'll simplify and use `p/thrown?` without a reader conditional.

jeaye: > In general I think of the possible exceptions thrown by a function as part of its public API. It's certainly something people write code against.

Thanks for making the change. Across this suite, the exception type is not an important aspect of what we're testing, since exception types are not portable. From a cross-dialect perspective, and a function behavior perspective, the only thing that matters is that a certain call throws.

I understand why you feel the desire for this, but it's part of our design philosophy for this repo to not have that so we can generalize test behavior better.

perrygeo: Given the goal is cross-platform compatibility, that makes perfect sense. Each host will almost by definition have its own unique error types.

I could see a case for enumerating the exceptions in a reader conditional by platform, but this would be a distant tertiary goal, after getting full coverage and platform support.

test/clojure/core_test/merge_with.cljc
dgr: @perrygeo , right, this is not the same. `second-arg` is being used as the merge function and it is overwriting the first arg with the second during the merge.

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,97 @@
...
+
+(when-var-exists
+ merge-with
+ (deftest test-merge-with
+   (testing "`merge-with`"
+
+     (testing "zero arg behavior"
+       #?(:cljs (is (nil? (merge-with)))
+          :clj  (is (p/thrown? (merge-with)))))
+
```
dgr, line 15: There's a missing case here, I think: `(merge-with second-arg)`

That is, no maps, just the required merge function present.

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,97 @@
...
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists] :as p]))
+
+(defn second-arg [_a b] b)
+
+(when-var-exists
+ merge-with
+ (deftest test-merge-with
+   (testing "`merge-with`"
+
+     (testing "zero arg behavior"
```
dgr, line 12: Let's be clear that there is no zero-arg case for the function legitimately. The doc string shows `f` as a required arg, so the CLJ case is throwing because it's short args.

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,97 @@
...
+       (is (= {:a 2} (merge-with second-arg {:a 1} {:a 2})))
+       (is (= {:a 3} (merge-with + {:a 1} {:a 2})))
+       (is (= {:a 3 :b 1} (merge-with + {:a 1} {:b 1} nil {:a 2})))
+       (is (= {:a [1 2] :b 3} (merge-with (fn [a b] (if (vector? a)
+                                                      (conj a b)
+                                                      [a b]))
+                                          {:a 1} {:a 2} {:b 3}))))
+
+     (testing "sorted"
+       (let [result (merge-with +
```
dgr, line 61: I would also try this where the first arg is a sorted map and the second is a regular hash map, to validate that the sorting behavior accrues with the first argument. I'd also test it with three maps, where the first is sorted to validate how the sorting behavior is propagated through the folding-in of the later maps.

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,97 @@
...
+       (is (= {:a 1} (merge-with second-arg {:a 1} nil {})))
+       (is (= {:a 1} (merge-with second-arg {:a 1} {} nil))))
+
+     (testing "f is applied left-to-right across maps"
+       ;; str is non-commutative, so order reveals the fold direction
+       (is (= {:a "123"} (merge-with str {:a "1"} {:a "2"} {:a "3"}))))
+
+     (testing "falsey values still trigger f on collision"
+       (is (= {:a false} (merge-with second-arg {:a true} {:a false})))
+       (is (= {:a nil} (merge-with second-arg {:a true} {:a nil}))))
```
dgr, line 30: I'm not sure what this is supposed to be testing.  The keys are both `:a` and the value doesn't participate in the decision to merge or not.

test/clojure/core_test/merge_with.cljc
```diff
@@ -0,0 +1,97 @@
...
+          :clj  (is (= {0.0 'double, 0.0M 'bigdecimal}
+                       (merge-with str {0.0 'double} {0.0M 'bigdecimal}))))
+       #?(:cljs (is (= {0 "longdouble"}
+                       (merge-with str {0 'long} {0.0 'double})))
+          :clj  (is (= {0 'long, 0.0 'double}
+                       (merge-with str {0 'long} {0.0 'double})))))
+
+     (testing "metadata progation, when keys are equal but not identical"
+       (is (= {:meta1 true}
+              (meta (ffirst (merge-with + {^:meta1 {} 1} {^:meta2 {} 2}))))))
```
dgr, line 88: This is good. I would add a test with metadata on the values as well and make sure that it's preserved through the merge and through the merge function.

dgr (changes requested):
Overall, looks good.

jeaye:
@perrygeo Any interest in iterating and getting this merged?

perrygeo:
Obviously, having addressed a dozen comments above, I have an interest in iterating! But not on someone else's schedule. If there is an urgency or a need for perfection prior to merge, count me out. Take my contribution or leave it, but please unassign me. Thanks!

jeaye:
> Obviously, having addressed a dozen comments above, I have an interest in iterating! But not on someone else's schedule. If there is an urgency or a need for perfection prior to merge, count me out. Take my contribution or leave it, but please unassign me. Thanks!

Ah, bad wording on my part, Matthew. I'm sorry to make you feel unappreciated. Yesterday, I was going through old/stale PRs and tickets and this was one of them. Yes, there's no urgency, but it's a reasonable thing to follow up on a PR after a month of inactivity to check if there's still interest and intention to get it merged. When that PR has outstanding comments, one also wants to know if there's still interest in iterating further.

I have unassigned you from the ticket, as you requested. I'll take this as feedback that our comments/reviews feel like they demand perfection, as well. If others feel similarly, that would make contributing quite stressful, so it's important to know.

PR #904, opened 2026-06-03
Add unicode char tests

### LINKED ISSUE #799: Add Unicode Test Cases to `clojure.core-test.char`
There is an outstanding `TODO` to add unicode test cases to `char` in `clojure.core-test.char`. I'm adding this issue to make sure it isn't lost.

E-A-Griffin:
This closes https://github.com/jank-lang/clojure-test-suite/issues/799

### DIFF
--- a/test/clojure/core_test/char.cljc
+++ b/test/clojure/core_test/char.cljc
@@ -17,7 +17,7 @@
        (is (= \ষ (char 2487))))
      (testing "4+ byte characters throw"
        (is #?(:jank    (= (first "𐅦") (char 65895))
-              ;; this seems to be a one off error
+              ;; this seems to be an off by one error
               :lpy     (= (first "𐅧") (char 65895))
               :cljs    (= \ŧ (char 65895))
               :default (p/thrown? (char 65895))))))
--- a/test/clojure/core_test/char.cljc
+++ b/test/clojure/core_test/char.cljc
@@ -1,5 +1,5 @@
 (ns clojure.core-test.char
-  (:require [clojure.test :as t :refer [are deftest is]]
+  (:require [clojure.test :as t :refer [are deftest is testing]]
             [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists] :as p]))
 
 (when-var-exists char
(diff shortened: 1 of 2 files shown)


test/clojure/core_test/concat.cljc
```diff
@@ -7,8 +7,10 @@
     (testing "arity 0"
       ;; returns an unrealized, empty lazy seq
       (let [s (concat)]
-        (is (p/lazy-seq? s))
-        (is (not (realized? s)))
+        (is #?(:lpy     (not (p/lazy-seq? s))
+               :default (p/lazy-seq? s)))
+        (is #?(:lpy     true
+               :default (not (realized? s))))
```
E-A-Griffin, line 13: These were failing for me locally and annoying me so just fixed them as part of this work, can separate this to a different PR if need be

test/clojure/core_test/char.cljc
```diff
@@ -9,9 +9,18 @@
...
-     )
+     \A     \A)
+   (testing "unicode"
+     (testing "2 byte characters are valid"
+       (is (= \¡ (char 161))))
+     (testing "3 byte characters are valid"
+       (is (= \ষ (char 2487))))
+     (testing "4+ byte characters throw"
+       (is #?(:jank    (= (first "𐅦") (char 65895))
+              ;; this seems to be an off by one error
```
jeaye, line 20: Is this a Basilisp bug? Shall we ping Chris?

E-A-Griffin: Seems like it

jeaye: @chrisrink10 heads up.

jeaye:
Thanks, Emma!

test/clojure/core_test/concat.cljc
chrisrink10: The fix for this was already merged into Basilisp's main branch, so you probably are not on the latest commit.

In the future, I would prefer if you either made sure you were up to date or just didn't run the tests locally at all.

cc @jeaye

test/clojure/core_test/char.cljc
chrisrink10: This function [delegates directly to Python `chr`](https://github.com/basilisp-lang/basilisp/blob/88c06799a1e0e9f96eaa23e41fbcf0ea02648243/src/basilisp/core.lpy#L1719). Are you sure Basilisp is the one in error?

test/clojure/core_test/concat.cljc
jeaye: Ok, yeah. Thanks for trying to keep up on the different dialects, Emma. It's too much to juggle, though, so we should just focus on the CI jobs that we're running and let dialect devs deal with everything else.

PR #907, opened 2026-06-12
Add tests for clojure.core/identity

### LINKED ISSUE #305: `identity`

brandoncorrea:
Closes https://github.com/jank-lang/clojure-test-suite/issues/305.

### `is` vs `are`

As commented in the code, some dialects seem to evaluate `are` parameters more than once, which is problematic for object equality. Test values are declared eagerly and tested using `doseq` / `is` instead.

### ##NaN

It turns out that most dialects don't consider `##NaN` to be `=` or `identical?` to itself. This case is tested separately with `NaN?`.

### DIFF
--- a/test/clojure/core_test/identity.cljc
+++ b/test/clojure/core_test/identity.cljc
@@ -30,13 +30,13 @@
         #uuid "f81d4fae-7dec-11d0-a765-00a0c91e6bf6"
         #inst "2010-11-12T13:14:15.666-05:00"
         (atom nil)))
-  
+
     (testing "NaN"
       (is (NaN? (identity ##NaN))))
-  
+
     (testing "Meta Preservation"
       (is (true? (:foo (meta (identity ^:foo {}))))))
-  
+
     (testing "Lazy Infinite Sequence"
       (let [infinite-seq (map inc (range))]
         (is (not (realized? (identity infinite-seq))))))))
--- a/test/clojure/core_test/identity.cljc
+++ b/test/clojure/core_test/identity.cljc
@@ -1,0 +1,35 @@
+(ns clojure.core-test.identity
+  (:require [clojure.test :refer [deftest is]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists]]))
+
+; Values for tests are declared eagerly, instead of using
+; clojure.test/are. Some dialects evaluate `are` params more
+; than once, causing false negatives with object equality.
+(def test-vals
+  [nil
+   true
+   false
+   ""
+   "foo"
+   \a
+   :foo
+   :foo.bar/baz
+   'foo
+   'foo.bar/baz
+   0
+   1
+   0.1
+   -0.1
+   ##Inf
+   ##-Inf
+   []
+   (list)
+   #uuid "f81d4fae-7dec-11d0-a765-00a0c91e6bf6"
+   #inst "2010-11-12T13:14:15.666-05:00"
+   (atom nil)])
+
+(when-var-exists identity
+  (deftest test-identity
+    (is (NaN? (identity ##NaN))) 
+    (doseq [value test-vals]
+      (is (identical? value (identity value))))))
--- a/test/clojure/core_test/identity.cljc
+++ b/test/clojure/core_test/identity.cljc
@@ -1,35 +1,42 @@
 (ns clojure.core-test.identity
-  (:require [clojure.test :refer [deftest is]]
+  (:require [clojure.test :refer [are deftest is testing]]
             [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists]]))
 
-; Values for tests are declared eagerly, instead of using
-; clojure.test/are. Some dialects evaluate `are` params more
-; than once, causing false negatives with object equality.
-(def test-vals
-  [nil
-   true
-   false
-   ""
-   "foo"
-   \a
-   :foo
-   :foo.bar/baz
-   'foo
-   'foo.bar/baz
-   0
-   1
-   0.1
-   -0.1
-   ##Inf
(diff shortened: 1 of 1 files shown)


test/clojure/core_test/identity.cljc
```diff
@@ -0,0 +1,35 @@
...
+   (list)
+   #uuid "f81d4fae-7dec-11d0-a765-00a0c91e6bf6"
+   #inst "2010-11-12T13:14:15.666-05:00"
+   (atom nil)])
+
+(when-var-exists identity
+  (deftest test-identity
+    (is (NaN? (identity ##NaN))) 
+    (doseq [value test-vals]
+      (is (identical? value (identity value))))))
```
NoahTheDuke: maybe put the value or the type of value as the docstring for the assertion?

NoahTheDuke: on second thought, i don't know if that's required...

brandoncorrea: Happy to add docstrings, but my hope is that failing assertion will document the value in the terminal output.

test/clojure/core_test/identity.cljc
```diff
@@ -0,0 +1,35 @@
+(ns clojure.core-test.identity
+  (:require [clojure.test :refer [deftest is]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists]]))
+
+; Values for tests are declared eagerly, instead of using
+; clojure.test/are. Some dialects evaluate `are` params more
```
jasalt: Huh, where exactly closure.test/are behaves oddly like that? Questioning why that wouldn’t be considered a bug to fix in a dialect (ticket could be linked here).

brandoncorrea: Actually, it seems all supported dialects fail the same way in these scenarios: jvm, cljs, cljr, bb, lpy, and phel.

```clojure
; fails with different object references
(deftest atoms-are=
  (are [x] (= x x)
    (atom nil)))

; fails with 3 instead of 1 (x referenced 3 times in do block)
(deftest increment-atom
  (let [counter (atom 0)]
    (are [x] (do x x x)
      (swap! counter inc))
    (is (= 1 @counter))))
```

jasalt: Interesting.

jeaye: If all dialects do the same thing, I don't think we can call it a failure. :slightly_smiling_face:

test/clojure/core_test/identity.cljc
```diff
@@ -0,0 +1,35 @@
...
+   -0.1
+   ##Inf
+   ##-Inf
+   []
+   (list)
+   #uuid "f81d4fae-7dec-11d0-a765-00a0c91e6bf6"
+   #inst "2010-11-12T13:14:15.666-05:00"
+   (atom nil)])
+
+(when-var-exists identity
```
jeaye, line 8: Let's please also add a test for meta preservation. i.e.

```clojure
(true? (:foo (meta (identity ^:foo {}))))
```

test/clojure/core_test/identity.cljc
```diff
@@ -0,0 +1,35 @@
...
+  (deftest test-identity
```
jeaye, line 9: Let's also add a test for lazy infinite sequences. We cannot check equality, but we can check that the value is not realized by passing it through `identity`.

jeaye (changes requested):
Thanks for the PR!

test/clojure/core_test/identity.cljc
```diff
@@ -0,0 +1,35 @@
...
+   'foo
+   'foo.bar/baz
+   0
+   1
+   0.1
+   -0.1
+   ##Inf
+   ##-Inf
+   []
+   (list)
```
E-A-Griffin: Nit: but I think `'()` would be more idiomatic

E-A-Griffin (commented):
Would be nice to test that calling `identity` on a side effecting expression multiple times just uses the realized value without side effect after the initial invocation, but not blocking

test/clojure/core_test/identity.cljc
jeaye: Also doesn't rely on the `list` var existing.

jeaye:
> Would be nice to test that calling `identity` on a side effecting expression multiple times just uses the realized value without side effect after the initial invocation, but not blocking

I like the out of the box thinking. This isn't one we want to do, though. This wouldn't test `identity`, since the effecting argument is evaluated before `identity` is called. With the way Clojure compilers work, all arguments are fully evaluated before the function is applied. We have to rely on things like this, when writing tests, otherwise we end up contorting in all shapes to test how the compiler is handling things, but that's not what we're aiming to test.

test/clojure/core_test/identity.cljc
brandoncorrea: Good points! Switched this over to `'()` in the last commit.

test/clojure/core_test/identity.cljc
brandoncorrea: Added!

test/clojure/core_test/identity.cljc
brandoncorrea: Added this

test/clojure/core_test/identity.cljc
brandoncorrea: Asked about this in Slack. Seems this behavior is known / expected, since the `are` uses the form a template expression.

https://clojurians.slack.com/archives/C03S1KBA2/p1781354480453579

jeaye:
As a side note, NaN is not equal to itself, by definition. That's not a Clojure thing it's an IEEE 754 thing.

jeaye:
Thanks!

PR #910, opened 2026-06-19
Add tests for edn/read-string

### LINKED ISSUE #908: `clojure.edn/read-string`

brandoncorrea:
This PR closes https://github.com/jank-lang/clojure-test-suite/issues/908.

These tests attempt to cover all behaviors according to [edn-format](https://github.com/edn-format/edn).

Many of these tests were inspired by [ClojureScript's reader tests](https://github.com/clojure/clojurescript/blob/master/src/test/cljs/cljs/reader_test.cljs).

Additional tests were added to flesh out all behaviors of the EDN spec, as well as some edge cases, negative scenarios, dialect-specific divergences, and some EDN extensions that some dialects implement and others do not (reader-conditionalized where applicable).

Babashka, JVM, CLR, and CLJS have all been tested and pass locally. Phel and Basilisp have been left to fail where apparent gaps exist.

### DIFF
--- a/test/clojure/edn_test/read_string.cljc
+++ b/test/clojure/edn_test/read_string.cljc
@@ -64,7 +64,7 @@
           "a\nb" "\"a\\nb\""
           "a\nb" "\"a\nb\""
           "a\nb\nc" "\"a\\nb\\nc\""
-          "a\nb\nc" "\"a\\nb\\nc\""
+          "a\nb\nc" "\"a\nb\nc\""
           "\n\\n\n" "\"\\n\\\\n\\n\""
           "\n\"\n" "\"\\n\\\"\\n\""
           "\ra\r" "\"\\ra\\r\""
--- a/test/clojure/edn_test/read_string.cljc
+++ b/test/clojure/edn_test/read_string.cljc
@@ -22,6 +22,9 @@
 
 (when-var-exists clojure.edn/read-string
   (deftest test-read-string
+    (testing "Throws Eval Reader"
+      (are-thrown (edn/read-string "#=(+ 1 2)")))
+
     (testing "nil and Booleans"
       (are-read-as
         nil "nil"
--- a/test/clojure/edn_test/read_string.cljc
+++ b/test/clojure/edn_test/read_string.cljc
@@ -24,13 +24,14 @@
   (deftest test-read-string
     (testing "nil and Booleans"
       (are-read-as
-        nil ""
         nil "nil"
         true "true"
         false "false"))
 
     (testing "Whitespace Only"
-      (are-read-as nil "  , ,, \t \r \n \r\n"))
+      (are-read-as
+        nil ""
+        nil "  , ,, \t \r \n \r\n"))
 
     (testing "Strings"
       (are-read-as
@@ -40,34 +41,22 @@
 
       (testing "Multi-line"
         (are-read-as
-          "\n" "\"\\n\""
+          "\n" "\"\n\""
           "\n\n" "\"\\n\\n\""
-          "\na" "\"\\na\""
-          "a\n" "\"a\\n\""
-          "a\nb" "\"a\\nb\""
-          "a\nb" "\"a\nb\""
-          "a\nb\nc" "\"a\\nb\\nc\""
-          "a\nb\nc" "\"a\nb\nc\""
-          "\n\\n\n" "\"\\n\\\\n\\n\""
-          "\n\"\n" "\"\\n\\\"\\n\""
-          "\ra\r" "\"\\ra\\r\""
-          "\r\na\r\n" "\"\\r\\na\\r\\n\""
+          "\r\n" "\"\r\n\""
+          "a\nb\nc" "\"a\\nb\nc\""
           '(def a "\n") "(def a \"\\n\")"))
 
       (testing "Unicode Escapes"
         (are-read-as
           "A" "\"\\u0041\""
           "AB" "\"\\u0041\\u0042\""
-          "→" "\"\\u2192\""
           "a→b" "\"a\\u2192b\""))
 
-      (testing "Unsupported Escapes"
-        (are-thrown "\"\\q\""))
-
-      (testing "Malformed Unicode Escapes"
+      (testing "Malformed Escapes"
         (are-thrown
-          "\"abc \\ua\""                                    ;; truncated
-          "\"abc \\x0z  ...etc\""                           ;; incorrect code
+          "\"\\q\""                                         ;; unsupported
+          "\"abc \\u000\""                                  ;; truncated
           "\"abc \\u0g00 ..etc\"")))                        ;; incorrect code
 
     (testing "Characters"
@@ -91,7 +80,7 @@
           (char 00) "\\o0"
           (char 0101) "\\o101"
           (char 0377) "\\o377")
-        ;; above the 377 maximum
+        ;; above 0377
         #?(;; cljs is lenient, reading a broader range of octal characters
            :cljs    (are-read-as
                       (char 0400) "\\o400"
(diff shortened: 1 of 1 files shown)


jeaye:
Thanks, Brandon!

@dgr What are your thoughts on the approach here?

test/clojure/edn_test/read_string.cljc
```diff
@@ -0,0 +1,901 @@
...
+(defn epoch-millis [date]
+  #?(:cljr    (.ToUnixTimeMilliseconds (DateTimeOffset. date))
+     :lpy     (* 1000 (.timestamp date))
+     :phel    (php/intval (php/-> date (format "Uv")))
+     :cljs    (.valueOf date)
+     :default (.getTime date)))
+
+(defn read-opts [opts s]
+  #?(:lpy     (edn/read-string s opts)
+     :phel    (edn/read-string s opts)
```
dgr: So, these dialects flip the arg order? I suppose they are using standard optional args to collect the options. That's a bug, IMO, and I wouldn't paper over it with a reader conditional.

test/clojure/edn_test/read_string.cljc
```diff
@@ -0,0 +1,901 @@
...
+
+    (testing "Whitespace Only"
+      (are-read-as-nil
+        " "
+        "  "
+        ","
+        ",,"
+        "\t"
+        "\r"
+        "\n"))
```
dgr: I'd probably add `"\r\n"` since that's what Windows systems will actually provide.

dgr (changes requested):
Overall, this is great. I would be careful about conditionalizing for other dialects. In some cases, some of these things seem like bugs and it's appropriate for the test to fail. Then leave it for the dialect authors to make the decision of what they want to do. If you conditionalize everything, it gives the dialect authors a false sense that everything is fine because the test passes and they don't review it.

dgr:
> @dgr What are your thoughts on the approach here?

I love this! Overall, this is quite well done and comprehensive.

test/clojure/edn_test/read_string.cljc
brandoncorrea: Added!

test/clojure/edn_test/read_string.cljc
brandoncorrea: Good point, this was one of the things I was a bit unsure on whether to keep or conditionalize. Removed it!

test/clojure/edn_test/read_string.cljc
```diff
@@ -0,0 +1,894 @@
...
+        'a1b2 "a1b2"
+        'foo "foo"
+        'foo/bar "foo/bar"
+        ;; basilisp fails compilation when using symbol literals on these three
+        (symbol "foo.bar" "baz.buzz") "foo.bar/baz.buzz"
+        (symbol "foo" "/") "foo//"
+        (symbol "foo.bar" "/") "foo.bar//"
+        'foo:bar "foo:bar"
+        ;; phel fails compilation with 'foo#bar
+        (symbol "foo#bar") "foo#bar"
```
brandoncorrea: @dgr what are your thoughts on cases like this? I'd like to use the literals, but in some cases, basilisp and phel reject them. Based on your earlier review, I imagine we'd want to use the literals and let these dialects fail?

jeaye: We should be writing tests for Clojure JVM, Clojure CLR, and babashka. Any failures should be picked up by the dialect owners when they run the test suite. It will be up to them to fix or workaround the failures. We do this so that test writers don't have to worry about all of the different dialect differences.

jeaye:
> > @dgr What are your thoughts on the approach here?
> 
> I love this! Overall, this is quite well done and comprehensive.

I agree, the tests are comprehensive. I also appreciate Brandon's work in getting these written. I want to convey caution before we continue, though. **Are we testing the right thing?** This repo is a test suite for core functions, to ensure they exist and behave correctly.

I would imagine that the `edn/read-string` should just be checking that we can read strings. Instead, we are thoroughly testing the underlying EDN reader. Let's scale this forward: when we add `read-json`, will we then implement and own a comprehensive JSON reader (and writer) testing suite? This is way beyond the scope of the repo.

As an example of how we **don't** do this, in other cases.

1. We don't test the underlying runtime, to ensure that calling fns with incorrect arg counts throws
2. We don't test the underlying compiler, to ensure that forms like `let`, `throw`, and `fn` behave correctly
3. We don't test the underlying runtime, to ensure that all possible code given to `eval` can be evaluated (see [here](https://github.com/jank-lang/clojure-test-suite/blob/main/test/clojure/core_test/eval.cljc))
4. We don't test the underlying runtime, to ensure that threads created by `future` are actually running in parallel

Overall, we don't test the underlying compiler+runtime. This is a unit testing repo for core functions. If we continue down this path, this repo will grow to contain:

1. Comprehensive unit tests for core functions
2. Comprehensive EDN reader/writer tests (`clojure.edn`)
3. Comprehensive JSON reader/writer tests (`clojure.data.json`)
5. Comprehensive XML reader/writer tests (`clojure.data.xml`)
6. Comprehensive CSV reader/writer tests (`clojure.data.csv`)
7. [And so on...](https://github.com/orgs/clojure/repositories?q=data)

Keep in mind that we have a policy against LLM-generated code. Outside of generating this test code, we cannot reasonably expand the scope of this repo to contain all of these tests. I think this would be biting off far more than we can chew.

Brandon asked if `clojure.edn/read-string` should be allowed in this repo and we said yes. However, the tests I had in mind for it were very much along the lines of the `eval` tests, not a comprehensive test of the entire reader. **I should have been more clear about that, out of respect for Brandon.** In order to not have helpful folks feel like their time is wasted, this is something we need to be very clear on.

brandoncorrea:
@jeaye I'd be happy to scale these tests back a bit. I was originally hesitant to create a 900 LOC PR, but went for it since each test seemed like it carried its weight.

I think we could certainly remove the _non-edn extension_ tests... since those are not part of the EDN spec. That should cut the size of the tests significantly by maybe 100 lines.
- Octal Escapes
- Octal, Hex, Radix Integers
- Ratios
- Namespaced Maps
- Metadata

Many of the other tests I think we could go without altogether or collapse to just a handful of cases. A few candidates I see just skimming through:
- Collapse some BigDecimal and Exponent cases
- Overflows could probably be removed
- Collapse Collections + Key Uniqueness cases
- Many of the Whitespace tests throughout could get collapsed

test/clojure/edn_test/read_string.cljc
dgr: Yep, as @jeaye says, use CLJ, CLJS, BB, and CLJR as your guide for what "normal" is. Assume deviations from that are issues with the dialect for the purpose of writing the test. They may or may not be, but it's up to the dialect authors to look at the failures and determine if they want to conditionalize the test to make it pass or whether they want to alter the implementation to make the test pass (possibly far in the future). What you don't want as the test author is to be making those decisions for the dialect author. Now, that said, there are real differences between CLJ and CLJS, for instance, so you can and should conditionalize those when it isn't a bug (and determining whether it is a bug or not might require some time on Clojurians Slack).

jeaye:
> @jeaye I'd be happy to scale these tests back a bit. I was originally hesitant to create a 900 LOC PR, but went for it since each test seemed like it carried its weight.

Yeah, this is 30x larger than the median test file in the repo (~30 lines). That is generally an indicator that EITHER an LLM was used and there's serious over-testing OR we're being way too specific about something.

For `clojure.edn/read-string`, I think the main things to test will be along these lines:

* Empty input => gives nil
* Single valid input => gives value
* Multiple valid inputs => gives first value?
* Invalid input => throws
* Unsafe reader macros are not evaluated (`edn/read-string` is supposed to be safe for untrusted inputs)

Then we can test each of the options to ensure their behavior works, but each of those only needs to test the happy path and any edge cases we can figure.

dgr:
I think it's quite reasonable to test `edn/read-string` to determine if it complies with the EDN behavior. To see why this is the case, think about the opposite. If `end/read-string` read a simple form but failed on the bulk of the valid EDN strings that we threw at it, wouldn't people immediately say, "This test is a joke and doesn't really test anything." Yes, there are other cases where we're not exhaustive. Perhaps we should be testing more in those cases. To the extent that we're testing anything in this test suite, I think we should do a good job of it. Some of the things that we don't test are because it's extremely difficult to test them (e.g., that a future is actually running in parallel). But to the extent that simple `is` tests can be written to validate behavior, we should try to validate more rather than less, up to the point of, but not including, completely over-testing and attempting to validate bug-for-bug behavior in out-of-bounds corner cases. So, I could see us doing something similar with JSON or XML reading.

That said, perhaps this is a progressive thing, where we shoot for a minimum simple test and if somebody wants to go further, more power to them. Perhaps we don't require such great tests as @brandoncorrea delivered, but we gratefully accept them if somebody wants to write them.

Finally, I think this also goes to the point I made to you privately on Slack, @jeaye. In these non-core libraries, if they are based mostly on Clojure code that will be reused in multiple implementations and the behavior is what it is as long as `core` is implemented correctly, then we don't need exhaustive tests. But exhaustive tests are useful in detecting issues with a library that relies on a lot of native platform code that must be reimplemented to port the library. In that case, the exhaustive tests validate those portions of the lib that needed to be rewritten.

test/clojure/edn_test/read_string.cljc
```diff
@@ -0,0 +1,894 @@
...
+        '$dollar "$dollar"
+        '_discard "_discard"
+        '<lt "<lt"
+        '>gt ">gt"
+        '.method ".method"
+        '.-attr ".-attr"))
+
+    (testing "Keywords"
+      (are-read-as
+        :1 ":1"
```
jeaye, line 122: Maybe a test for just `:`?

jeaye (changes requested):
Thank you, Brandon!

jeaye:
Following up publicly, after speaking privately with folks. We're going to move forward to merge Brandon's hard work, which intentionally grows the scope of the clojure-test-suite to now handle testing parsers for data formats. We're going to trim down some of the over-testing prior to merging.

test/clojure/edn_test/read_string.cljc
brandoncorrea: Awesome, this is good info. I'll be pushing changes soon that rip out pretty much all phel/basilisp conditionals.

test/clojure/edn_test/read_string.cljc
brandoncorrea: Added into _Invalid Tokens_ in the latest commit!

test/clojure/edn_test/read_string.cljc
```diff
@@ -0,0 +1,894 @@
...
+        '(4) "(#_(1 2 3) 4)"
+        '(4) "(#_[1 2 3] 4)"
+        '(2) "(#_{:a 1} 2)"
+        '(3) "(#_#{1 2} 3)"
+        {:a 1} "{:a;foo:b\n1}"
+        {:a 1 :c 3} "{:a 1;:b 2\n:c 3}"
+        {:a 1 :c 3} "{:a 1 #_:b #_2 :c 3}"
+        {:a 1} "{:a #_nope 1}"
+        {:a 2} "{:a #_,\t\n\r 1 2}"
+        {:a 1 :c 3} "{:a 1 #_:b #_2 :c 3}"
```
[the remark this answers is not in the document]
brandoncorrea: Removed the duplicate in the latest commit

brandoncorrea:
Trimmed down to ~500 lines and addressed some of the earlier feedback. I think the tests are in a better place now and the changes ready for review.

jeaye:
Great work, Brandon!

PR #913, opened 2026-06-21
Add tests for `clojure.string/replace`

### LINKED ISSUE #457: `replace`

djblue:
This PR closes https://github.com/jank-lang/clojure-test-suite/issues/457.

### DIFF
--- a/test/clojure/string_test/replace.cljc
+++ b/test/clojure/string_test/replace.cljc
@@ -1,0 +1,19 @@
+(ns clojure.string-test.replace
+  (:require [clojure.string :as str]
+            [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists] :as p]))
+
+(when-var-exists str/replace
+  (deftest test-replace
+    (is (p/thrown? (str/replace nil "")) "Input must be string")
+
+    (is (= "" (str/replace "x" "x" "")) "Replace can empty strings")
+    (is (= "" (str/replace "" "" "")) "Check for infinite loops")
+    (is (= "yxy" (str/replace "x" "" "y")) "Empty string matches between all characters")
+    (is (= "yyxyy" (str/replace "x" "" "yy")) "Empty string matches between all characters")
+
+    (is (= "yy" (str/replace "xx" "x" "y")))
+    (is (= "y" (str/replace "xx" "xx" "y")))
+    (is (= "xx" (str/replace "xx" "xxx" "y")))
+    (is (= "yyy" (str/replace "yxy" "x" "y")))))
+
--- a/test/clojure/string_test/replace.cljc
+++ b/test/clojure/string_test/replace.cljc
@@ -7,8 +7,11 @@
   (deftest test-replace
     (is (p/thrown? (str/replace nil "x" "y")) "Input must be string")
 
-    (is (= "" (str/replace "x" "x" "")) "Replace can empty strings")
-    #?(:cljr "ClojureCLR doesn't support empty strings for match"
+    (is (= "" (str/replace "x" "x" "")) "Replace can produce an empty string")
+    (is (= "" (str/replace "" "x" "x")) "Empty string has nothing to replace")
+    (is (= "" (str/replace "" "x" "")) "Empty string has nothing to replace")
+    #?(:cljr
+       (is true "ClojureCLR doesn't support empty strings for match")
        :default
        (do (is (= "" (str/replace "" "" "")) "Check for infinite loops")
            (is (= "yxy" (str/replace "x" "" "y")) "Empty string matches between all characters")
@@ -17,5 +20,6 @@
     (is (= "yy" (str/replace "xx" "x" "y")))
     (is (= "y" (str/replace "xx" "xx" "y")))
     (is (= "xx" (str/replace "xx" "xxx" "y")))
-    (is (= "yyy" (str/replace "yxy" "x" "y")))))
-
+    (is (= "yyy" (str/replace "yxy" "x" "y")))
+    (is (= "yyx" (str/replace "xxxxx" "xx" "y")))
+    (is (= "yyyyy" (str/replace "xyyyx" "x" "y")))))
--- a/test/clojure/string_test/replace.cljc
+++ b/test/clojure/string_test/replace.cljc
@@ -86,6 +86,10 @@
          (is (= (str `bar) (str/replace `foo "foo" "bar")))
          (is (= (str [`bar]) (str/replace [`foo] "foo" "bar")))))
 
+    (testing "Unicode Emoji"
+      (is (= "😀" (str/replace "☹️" "☹️" "😀")))
+      (is (= "😀😀😀" (str/replace "☹️☹️☹️" "☹️" "😀"))))
+
     #?(:cljs nil ;; cljs converts chars into strings
        :default
        (testing "Invalid parameter combinations"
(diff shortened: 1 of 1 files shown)


test/clojure/string_test/replace.cljc
```diff
@@ -0,0 +1,21 @@
...
+(when-var-exists str/replace
+  (deftest test-replace
+    (is (p/thrown? (str/replace nil "x" "y")) "Input must be string")
+
+    (is (= "" (str/replace "x" "x" "")) "Replace can empty strings")
+    #?(:cljr "ClojureCLR doesn't support empty strings for match"
+       :default
+       (do (is (= "" (str/replace "" "" "")) "Check for infinite loops")
+           (is (= "yxy" (str/replace "x" "" "y")) "Empty string matches between all characters")
+           (is (= "yyxyy" (str/replace "x" "" "yy")) "Empty string matches between all characters")))
```
jeaye: Some cases come to mind, just looking at this permutation visually.

1. arg 1: empty, arg 2: non-empty, arg 3: non-empty
2. arg 1: non-empty, arg 2: non-empty, arg 3: empty
3. arg 1: empty, arg 2: non-empty, arg 3: empty

jeaye (changes requested):
Thanks, Chris!

test/clojure/string_test/replace.cljc
```diff
@@ -0,0 +1,25 @@
+(ns clojure.string-test.replace
+  (:require [clojure.string :as str]
+            [clojure.test :as t :refer [deftest testing is are]]
+            [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists] :as p]))
+
+(when-var-exists str/replace
+  (deftest test-replace
```
dgr, line 30: One thing I've found useful is to take the doc string for the function under test and insert it as a comment at the top of the file so that it's easy for reviewers and others looking at the file later to see what the function "guarantees" for its behavior. Not a biggie, but would be useful.

test/clojure/string_test/replace.cljc
```diff
@@ -0,0 +1,25 @@
...
+    (is (p/thrown? (str/replace nil "x" "y")) "Input must be string")
```
dgr: Move the exceptional cases toward the end. Start with the happy path first, validating that the function does what the doc string claims it does. Then move onto the exceptional cases.

test/clojure/string_test/replace.cljc
```diff
@@ -0,0 +1,25 @@
...
+
+    (is (= "" (str/replace "x" "x" "")) "Replace can produce an empty string")
+    (is (= "" (str/replace "" "x" "x")) "Empty string has nothing to replace")
+    (is (= "" (str/replace "" "x" "")) "Empty string has nothing to replace")
+    #?(:cljr
+       (is true "ClojureCLR doesn't support empty strings for match")
```
dgr: This is probably a bug and should be reported to the CLJR team. @dmiller , can you look at this? Seems like empty strings should be part of the contract.

test/clojure/string_test/replace.cljc
```diff
@@ -0,0 +1,25 @@
...
+       :default
+       (do (is (= "" (str/replace "" "" "")) "Check for infinite loops")
```
dgr: Nicely done!

test/clojure/string_test/replace.cljc
```diff
@@ -0,0 +1,25 @@
...
+           (is (= "yxy" (str/replace "x" "" "y")) "Empty string matches between all characters")
+           (is (= "yyxyy" (str/replace "x" "" "yy")) "Empty string matches between all characters")))
+
+    (is (= "yy" (str/replace "xx" "x" "y")))
+    (is (= "y" (str/replace "xx" "xx" "y")))
+    (is (= "xx" (str/replace "xx" "xxx" "y")))
+    (is (= "yyy" (str/replace "yxy" "x" "y")))
+    (is (= "yyx" (str/replace "xxxxx" "xx" "y")))
+    (is (= "yyyyy" (str/replace "xyyyx" "x" "y")))))
```
dgr: Additional case: overlapping replacements. For instance:
```
(is (= "fooyx" (str/replace "xyxyx" "xyx" "foo")))
```
The first replacement is performed but the second cannot be because the string has been altered.

Also:
```
(is (= "xyz" (str/replace "y" "y" "xyz")))
```

So, the `y` that is part of the replacement is not subject to further replacement.

dgr (changes requested):
Overall, great job. Some tweaks and additional cases requested.

dgr:
Spoke too soon. Shame on me for not having read the doc string for `clojure.string/replace`. There are a lot of cases that are not covered. Here's the doc string:

```
(replace s match replacement)

Replaces all instance of match with replacement in s.
  match/replacement can be:
  string / string
 char / char
 pattern / (string or function of match).
  See also replace-first.
  The replacement is literal (i.e. none of its characters are treated
 specially) for all cases above except pattern / string.
  For pattern / string, $1, $2, etc. in the replacement string are
 substituted with the string that matched the corresponding
 parenthesized group in the pattern.  If you wish your replacement
 string r to be used literally, use (re-quote-replacement r) as the
 replacement argument.  See also documentation for
 java.util.regex.Matcher's appendReplacement method.
  Example:
 (clojure.string/replace "Almost Pig Latin" #"\b(\w)(\w+)\b" "$2$1ay")
 -> "lmostAay igPay atinLay"
```

Note that the match and the match/replacement can be a string/string, char/char, or pattern/string or function.

So, those char/char and pattern/string or function cases need to be included.

Note that the doc string includes a test case that you should probably include in the test.

dgr:
> I'm only focusing only on strings currently, but will add regex examples in a follow up PR. Let me know if you have more interesting edge cases to include.

Apologies. Just read this. Can we get all the test cases? I'm hesitant to include half the tests and come back to it later because later frequently gets lost, and then the presence of a test suggests that the function is being fully tested. Further, an open issue with a test in the code is confusing (is it done or not?). Yes, someone can figure it out, but it's not straightforward.

So, @jeaye, I'd recommend we not merge this until we get all of it (ignoring mistakes of omission).

jeaye:
> > I'm only focusing only on strings currently, but will add regex examples in a follow up PR. Let me know if you have more interesting edge cases to include.
> 
> Apologies. Just read this. Can we get all the test cases? I'm hesitant to include half the tests and come back to it later because later frequently gets lost, and then the presence of a test suggests that the function is being fully tested. Further, an open issue with a test in the code is confusing (is it done or not?). Yes, someone can figure it out, but it's not straightforward.
> 
> So, @jeaye, I'd recommend we not merge this until we get all of it (ignoring mistakes of omission).

Yep, I agree.

@djblue We don't need to exhaustively test the regex cases. We're not testing the underlying regex engine. We're just checking that regex can be used for matching, as well as edge cases like empty regex.

If we need to hold this open until someone can jump in and finish it, that's ok.

test/clojure/string_test/replace.cljc
E-A-Griffin: slight nit but `do` feels odd here? not a huge deal but might be better to use something with `@` like in https://github.com/jank-lang/clojure-test-suite/blob/main/test/clojure/core_test/max.cljc#L39-L51 (you can then use `nil` for the `cljr` form)

E-A-Griffin (commented):
Unicode string tests (i.e. strings with characters not covered by the ASCII standard set) would be nice, I could see that being a source of bugs (same for characters)

djblue:
@dgr, @E-A-Griffin how do you feel about the current state of the PR? Do you think this PR needs more work, or has it reached a point where it can be merged? I think I addressed all of the feedback, please let me know if I missed something.

jeaye (approved):
I'm pushing this through, since it works for me and we haven't heard back from Dave/Emma.

PR #914, opened 2026-06-22
Add test for clojure.core/run!

### LINKED ISSUE #470: `clojure.core/run!`

brandoncorrea:
This PR closes https://github.com/jank-lang/clojure-test-suite/issues/470.

Passes on Babashka, JVM, CLR, CLJS, and Basilisp.

### Phel

Three tests fail on Phel:
- _Collection Arguments > Maps_ (2 cases)
- _Terminates on Reduced_

The _Maps_ failures seem to be a real gap between Phel and the other dialects.

I would like some guidance on the _Terminates on Reduced_ test. This basically asserts that `run!` is implemented in terms of `reduce`, which the docstring claims:

> Runs the supplied procedure (via reduce), for purposes of side
> effects, on successive items in the collection. Returns nil

However, I wonder if this is documented more as an implementation detail rather than a contract.

### DIFF
--- a/test/clojure/core_test/run_bang.cljc
+++ b/test/clojure/core_test/run_bang.cljc
@@ -1,0 +1,84 @@
+(ns clojure.core-test.run-bang
+  (:require [clojure.core-test.portability #?(:cljs :refer-macros :default :refer) [when-var-exists] :as p]
+            [clojure.test :refer [are deftest is testing]]))
+
+(when-var-exists run!
+
+  (defn inspect-run! [proc init coll]
+    (let [result (volatile! init)]
+      (run! (fn [v] (vswap! result proc v)) coll)
+      @result))
+
+  (deftest test-run!
+    (testing "Always Nil"
+      (are [coll] (nil? (run! identity coll))
+        []
+        [1]
+        [1 2 3]
+        ["foo" "bar"]))
+
+    (testing "Causes Side-Effects"
+      (let [calls     (volatile! 0)
+            inc-calls (fn [_] (vswap! calls inc))]
+        (run! inc-calls [])
+        (is (zero? @calls))
+
+        (run! inc-calls [0])
+        (is (= 1 @calls))
+
+        (run! inc-calls [0 0])
+        (is (= 3 @calls))))
+
+    (testing "Collection Argument"
+      (are [coll] (= 6 (inspect-run! + 0 coll))
+        [1 2 3]
+        '(1 2 3)
+        #{1 2 3})
+
+      (testing "Nil"
+        (is (zero? (inspect-run! + 0 nil))))
+
+      (testing "Strings"
+        (are [expected s] (= expected (inspect-run! conj [] s))
+          [] ""
+          [\a] "a"
+          [\f \o \o] "foo"))
+
+      ;; phel passes values to the proc (other dialects pass [k v] entries)
+      (testing "Maps"
+        (are [expected m] (= expected (inspect-run! conj #{} m))
+          #{} {}
+          #{[:foo :bar]} {:foo :bar}
+          #{[:foo :bar] [:baz :buzz]} {:foo :bar :baz :buzz}))
+
+      (testing "Non-Seqable Values"
+        (are [v] (p/thrown? (run! identity v))
+          ;; basilisp, phel, and cljs interpret chars as strings
+          #?@(:lpy [] :phel [] :cljs [] :default [\a])
+          true
+          #uuid "00000000-0000-0000-0000-000000000000"
+          1
+          1.0
+          :foo
+          'foo)))
+
+    (testing "Passes Collection Sequentially"
+      (let [coll [:foo "bar" 'baz]]
+        (is (= coll (inspect-run! conj [] coll)))))
+
+    (testing "Terminates on Exception"
+      (let [calls (volatile! 0)
+            boom! (fn [_]
+                    (vswap! calls inc)
+                    (throw (ex-info "Boom!" {})))]
+        (is (p/thrown? (run! boom! (range 2))))
+        (is (= 1 @calls))))
+
+    ;; phel implements run! in terms of doseq (other dialects use reduce)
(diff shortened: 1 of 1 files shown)


jeaye:
> > Runs the supplied procedure (via reduce), for purposes of side
> > effects, on successive items in the collection. Returns nil
> 
> However, I wonder if this is documented more as an implementation detail rather than a contract.

Clojure's doc strings are contract, based on my discussions with the core team. If it's specified, in the doc string, it's guaranteed. If it's not specified, all bets are off.

test/clojure/core_test/run_bang.cljc
```diff
@@ -0,0 +1,84 @@
...
+        (is (= 3 @calls))))
+
+    (testing "Collection Argument"
+      (are [coll] (= 6 (inspect-run! + 0 coll))
+        [1 2 3]
+        '(1 2 3)
+        #{1 2 3})
+
+      (testing "Nil"
+        (is (zero? (inspect-run! + 0 nil))))
```
jeaye: What're we trying to test here, with respect to `run!`? This seems like we're testing that the `inspect-run!` machinery works and that `run!` doesn't do anything with `nil`, but I think we could simplify that to group it above like so:

```clojure
        (run! inc-calls nil)
        (is (zero? @calls))

        (run! inc-calls [])
        (is (zero? @calls))
```

Does that achieve the same goal?

test/clojure/core_test/run_bang.cljc
```diff
@@ -0,0 +1,84 @@
...
+
+      (testing "Strings"
+        (are [expected s] (= expected (inspect-run! conj [] s))
+          [] ""
+          [\a] "a"
+          [\f \o \o] "foo"))
```
jeaye: I have a similar concern here. For this block, we're testing that strings are sequences and the `conj` can append characters into vectors. I don't think any of this is specific to `run!`.

test/clojure/core_test/run_bang.cljc
```diff
@@ -0,0 +1,84 @@
...
+
+    (testing "Terminates on Exception"
+      (let [calls (volatile! 0)
+            boom! (fn [_]
+                    (vswap! calls inc)
+                    (throw (ex-info "Boom!" {})))]
+        (is (p/thrown? (run! boom! (range 2))))
+        (is (= 1 @calls))))
+
+    ;; phel implements run! in terms of doseq (other dialects use reduce)
```
jeaye: Let's leave it up to dialects to fix their own stuff, without encoding it as a comment. That way, when they fix it, we don't have an out of date comment. Also, that way you don't need to worry about dialects like Phel when writing tests.

The same thing applies to jank, which is intentionally omitted so that it's not intrusive. :)

test/clojure/core_test/run_bang.cljc
```diff
@@ -0,0 +1,84 @@
...
+      (let [coll [:foo "bar" 'baz]]
+        (is (= coll (inspect-run! conj [] coll)))))
+
+    (testing "Terminates on Exception"
+      (let [calls (volatile! 0)
+            boom! (fn [_]
+                    (vswap! calls inc)
+                    (throw (ex-info "Boom!" {})))]
+        (is (p/thrown? (run! boom! (range 2))))
+        (is (= 1 @calls))))
```
jeaye, line 50: Nice. This is a good one.

test/clojure/core_test/run_bang.cljc
```diff
@@ -0,0 +1,84 @@
...
+          true
+          #uuid "00000000-0000-0000-0000-000000000000"
+          1
+          1.0
+          :foo
+          'foo)))
+
+    (testing "Passes Collection Sequentially"
+      (let [coll [:foo "bar" 'baz]]
+        (is (= coll (inspect-run! conj [] coll)))))
```
jeaye: This is a good one.

test/clojure/core_test/run_bang.cljc
```diff
@@ -0,0 +1,84 @@
...
+          [] ""
+          [\a] "a"
+          [\f \o \o] "foo"))
+
+      ;; phel passes values to the proc (other dialects pass [k v] entries)
+      (testing "Maps"
+        (are [expected m] (= expected (inspect-run! conj #{} m))
+          #{} {}
+          #{[:foo :bar]} {:foo :bar}
+          #{[:foo :bar] [:baz :buzz]} {:foo :bar :baz :buzz}))
```
jeaye: This is specific to how map entries are processed, which is unrelated to `run!`. I don't think it's improving our `run!` coverage.

jeaye (changes requested):
Thank you, Brandon! I've left some comments along the lines of focusing on `run!` and not the implementation details of various sequences. Happy to discuss if it helps to clarify anything.

test/clojure/core_test/run_bang.cljc
brandoncorrea: Good point. Now that you bring it up, I wonder if we even need any of these "variable collection" sections. They effectively just test that the collections are seqable, which is unrelated to `run!`, as you mentioned.
- _Nil_
- _Strings_
- _Maps_

On the same note, I wonder if the _Non-Seqable Values_ section could be reduced to one test (such as below), rather than all the non-seqable types it currently tests.

```clojure
(is (p/thrown? (run! identity true)))
```

test/clojure/core_test/run_bang.cljc
brandoncorrea: Comments removed!

test/clojure/core_test/run_bang.cljc
brandoncorrea: This has been removed in the latest commit

test/clojure/core_test/run_bang.cljc
brandoncorrea: I like this - committed!

jeaye:
Nice job!

PR #917, opened 2026-06-26
Add `sleep` for jank

djblue:


test/clojure/core_test/portability.cljc
```diff
@@ -29,7 +30,10 @@
       :cljs #(js/setTimeout identity %)
       :clj Thread/sleep
       :lpy time/sleep
-      :phel #(phel.async/delay (/ % 1000)))
+      :phel #(phel.async/delay (/ % 1000))
+      :jank #(cpp/std.this_thread.sleep_for
```
djblue: I think having to make a lambda here is a bit awkward, and I'm down to update this reader conditional so all paths deal with ms directly, but I didn't want to refactor without discussion first.

jeaye: Sure, let's do that. Thanks for the offer.

jeaye:
Thanks!

PR #918, opened 2026-06-26
Fix `clojure.core-test.dissoc` for jank

djblue:


jeaye (approved):
That's a clean fix. Thanks!

PR #922, opened 2026-06-30
Add tests for clojure.core/every?

### LINKED ISSUE #261: `clojure.core/every?`

### LINKED ISSUE #375: `not-every?`

brandoncorrea:
This PR closes https://github.com/jank-lang/clojure-test-suite/issues/261

### Edit

This PR also closes https://github.com/jank-lang/clojure-test-suite/issues/375, `not-every?`

### DIFF
--- a/test/clojure/core_test/every_qmark.cljc
+++ b/test/clojure/core_test/every_qmark.cljc
@@ -4,59 +4,62 @@
 
 (defn boom! [x] (throw (ex-info "Boom!" {:x x})))
 
+(defn tests [every-fn]
+  (testing "Basic Predicates and Collections"
+    (is (false? (every-fn odd? [1 2 3])))
+    (is (true? (every-fn odd? [1 3 5])))
+    (is (false? (every-fn even? #{1 2 3})))
+    (is (true? (every-fn even? #{2 4 6}))))
+
+  (testing "Empty Collections"
+    (are [coll] (true? (every-fn boom! coll))
+      nil
+      []
+      '()))
+
+  (testing "Maps and Sets as Predicates"
+    (are [expected pred] (= expected (every-fn pred [:a :b :c]))
+      true #{:a :b :c}
+      false #{:a :c}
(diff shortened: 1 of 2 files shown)


brandoncorrea:
I noticed `not-every` https://github.com/jank-lang/clojure-test-suite/issues/375 is open as well, which is just `(not (every? ...))`.

Is the expectation for `(not (foo ...))` functions that we create the `foo` tests in `foo.cljc` and reference them in both `foo` and `not-foo` tests? I see that being done with `=` and `not=`.
- https://github.com/jank-lang/clojure-test-suite/blob/main/test/clojure/core_test/not_eq.cljc
- https://github.com/jank-lang/clojure-test-suite/blob/main/test/clojure/core_test/eq.cljc

If so, it wouldn't be much more work to include that as part of this PR.

jeaye:
> I noticed `not-every` #375 is open as well, which is just `(not (every? ...))`.
> 
> Is the expectation for `(not (foo ...))` functions that we create the `foo` tests in `foo.cljc` and reference them in both `foo` and `not-foo` tests? I see that being done with `=` and `not=`.
> 
>     * https://github.com/jank-lang/clojure-test-suite/blob/main/test/clojure/core_test/not_eq.cljc
> 
>     * https://github.com/jank-lang/clojure-test-suite/blob/main/test/clojure/core_test/eq.cljc
> 
> 
> If so, it wouldn't be much more work to include that as part of this PR.

Sure, I think this is ok. No need to duplicate everything.

jeaye (approved):
These look great to me!

E-A-Griffin (commented):
Mostly looks good! Can we add a case (or cases) where the `coll` (i.e. the second arg) is `nil`?

E-A-Griffin (approved):
Ignore my initial comment, I checked again and I see you already have it covered

jeaye:
Sorry for the delay. Please request a review to get my attention on these things.

PR #923, opened 2026-07-04
Fix big decimal str for jank

djblue:


test/clojure/core_test/str.cljc
```diff
@@ -58,7 +58,15 @@
...
            "-1"         -1.0M]
-          
+
+          :jank
+          ["0"          0/2
+           "1/2"        1/2
+           "-1/2"       -1/2
+           "0"          0.0M
+           "1"          1.0M
+           "-1"         -1.0M]
```
jeaye, line 68: This is a bug. We should be printing these just as the `:default` case shows.

PR #928, opened 2026-07-11
Improve `first` tests

dgr:
Add cases for arrays, maps, sets, and strings.


jeaye (approved):
Thanks for the improvements!

PR #932, opened 2026-07-11
Improve `contains?` tests

dgr:


test/clojure/core_test/contains_qmark.cljc
```diff
@@ -1,24 +1,110 @@
...
+      true {nil :a} nil                 ; doesn't matter if key is nil
+      true {:a nil} :a                  ; doesn't matter if val is nil
+      true {[:a 1] 17} [:a 1]           ; keys can be collections
+      false {[:a 1] 17} [:a 2]
+      true (sorted-map :a 1 :b 2) :a    ; no difference if sorted maps
+      false (sorted-map :a 1 :b 2) :c
+      true (sorted-map nil 1) nil
+      false {} :a
+      false {} nil
+      false nil :a
```
jeaye, line 33: Might be a nice idea to include a map entry check.

dgr: How so, exactly? Are you saying to treat a map entry like an indexed collection (e.g., a vector)?

For example:
```clojure
(contains? (first {:a 1}) 0)
;; => true
```

dgr: OK, I added tests both for map entries as well as metadata on collection keys of associative collections (maps and sets).

jeaye: Sorry for the delay. The tests you added for map entries are good, but not what I was thinking. The case I was considering was to check if a map contains a map entry. Something like:

```clojure
(let [m1 {:a 1} m2 {(first m1) true}]
  (is (false? (contains? m1 (first m1))))
  (is (contains? m2 (first m1))))
```

I think this is an interesting case, since I can see how checking if a map contains a map entry could work. This is the sort of thing which might accidentally work, for some runtimes.

dgr: Ah, I see what you're saying. OK, let me add that. Good catch.

dgr: Done

jeaye (approved):
Nice work, Dave!

PR #933, opened 2026-07-12
Improve `bigdec` tests

dgr:


test/clojure/core_test/bigdec.cljc
```diff
@@ -1,28 +1,53 @@
...
+        0M 0.0
+        -1M -1.0
+        0M "0"
+        1M "1"
+        1M "+1"                         ; explicit plus sign
+        -1M "-1"
+        0.5M "0.5"
+        -0.5M "-0.5"
+        1E+10M "1e10"                   ; exponential notation
+        1E+10M "1E10"
```
jeaye, line 25: Scientific notation also supports `+` and `-` prefixes. Probably worth adding.

dgr: +/- prefixes for the exponent or the mantissa, or both?

dgr: I just added all combinations. See if that works for you.

jeaye (approved):
Nice!

PR #934, opened 2026-07-12
Suppress sorted collection tests for Basilisp

chrisrink10:
Suppress a couple of sorted collection assertions in new tests for Basilisp.


jeaye (approved):
Works for me!

PR #935, opened 2026-07-12
Update Basilisp portability macro to use `assert-expr`

chrisrink10:
I merged updates to `basilisp.test` to support `assert-expr` and `do-report` in https://github.com/basilisp-lang/basilisp/pull/1345 so I wanted to update the portability macro here to use those now instead of the older non-standard utilities.


jeaye (approved):
Thanks!

PR #939, opened 2026-07-16
Add let-go Unicode scalar char expectation

nnunley:
Summary:
- add an :lg reader branch for char 65895 as Unicode scalar U+10167
- clarify Unicode-scalar behavior versus JVM UTF-16 char behavior
- preserve all existing dialect expectations

Validation:
- nooga/let-go char fixture: 9 passed, 0 failed, 0 errors

This lets let-go consume the shared fixture without a local skip while retaining its broad Unicode character support.


jeaye (approved):
Thank you!

PR #940, opened 2026-07-21
Improve `nth` tests

dgr:


test/clojure/core_test/nth.cljc
```diff
@@ -1,30 +1,156 @@
...
+
+        ;; arrays
+        0 (int-array [0 1 2 3 4 5 6 7 8 9]) 0
+        5 (int-array [0 1 2 3 4 5 6 7 8 9]) 5
+        9 (int-array [0 1 2 3 4 5 6 7 8 9]) 9
+
+        ;; Surprisingly, `nil` collection return `nil`, regardless of
+        ;; `index`
+        nil nil 0
+        nil nil 10)
```
jeaye: Do we want a `nil` collection with a negative index test?

dgr: Yep. We can. And I thought about it. But how much out of bounds testing do you want to do? There are tests with an actual collection with a -1 index at the bottom of the file in the exceptions section. To be clear, if it was me, I would test all this crazy stuff, and I think it should all throw an exception. But I know that slows everything down, and so Clojure was written in most cases to only deal with the happy path and you're on your own if you stray from that path. So, I've sort of trained myself to try to stick to the testing the happy path as well.

So, let me ask you this. Without consulting a REPL, what should `(nth nil -1)` do?

test/clojure/core_test/nth.cljc
```diff
@@ -1,30 +1,156 @@
...
+      ;; re-matchers are stateful, so set it up and then force the
+      ;; matching with `re-find`.
+      #?(:cljs nil ; CLJS doesn't have re-matcher
+         :default (let [m (-> (re-matcher #"(\d+),(\d+),(\d+)" "123,456,789")
+                              re-find)]
+                    (are [expected coll index] (= expected (nth coll index))
+                      "123,456,789" m 0
+                      "456" m 2
+                      "789" m 3)
+                    (is (p/thrown? (nth m 10)))))
```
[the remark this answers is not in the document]
dgr: Good catch. Fixed and pushed.

test/clojure/core_test/nth.cljc
```diff
@@ -1,30 +1,156 @@
...
+      ;; re-matchers are stateful, so set it up and then force the
+      ;; matching with `re-find`.
+      #?(:cljs nil ; CLJS doesn't have re-matcher
+         :default (let [m (-> (re-matcher #"(\d+),(\d+),(\d+)" "123,456,789")
+                              re-find)]
+                    (are [expected coll index] (= expected (nth coll index))
+                      "123,456,789" m 0
+                      "456" m 2
+                      "789" m 3)
+                    (is (= :default (nth m 10 :default)))))
```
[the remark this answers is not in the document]
dgr: Fixed and pushed.

test/clojure/core_test/nth.cljc
```diff
@@ -1,30 +1,156 @@
...
+        \l "hello" 2
+        \o "hello" 4
+
+        ;; arrays
+        0 (int-array [0 1 2 3 4 5 6 7 8 9]) 0
+        5 (int-array [0 1 2 3 4 5 6 7 8 9]) 5
+        9 (int-array [0 1 2 3 4 5 6 7 8 9]) 9
+
+        ;; Surprisingly, `nil` collection return `nil`, regardless of
+        ;; `index`
```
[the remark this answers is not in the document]
dgr: Fixed

test/clojure/core_test/nth.cljc
jeaye: > Without consulting a REPL, what should (nth nil -1) do?

Throw.

dgr: Except it doesn't throw. It returns `nil` because `nth` special cases `nil` such that if the collection is `nil`, it ignores the `index` and returns `nil`.

dgr: I added that case. It's odd behavior, IMO, but according to this [Ask Clojure record](https://ask.clojure.org/index.php/15195/nth-nil-10-returns-nil-when-it-should-throw?state=comment-15196&show=15196#c15196), it's correct.

dgr:
Should be ready to go now.

jeaye (approved):
Thanks, Dave!

PR #950, opened 2026-08-09
Update tests for `jank`

shantanu-sardesai:


test/clojure/core_test/key.cljc
```diff
@@ -14,13 +14,14 @@
...
         0
-        '()
-        '(1 2)
-        {}
-        {1 2}
-        []
-        [1 2]
-        #{}
-        #{1 2}))))
+        #?@(:jank [] ; jank doesn't have a concept of a map entry.
```
jeaye, line 18: I would clarify this to say that jank just returns `first`, since it assumes a map entry is a vector of two objects.

Same with `val`, except we use `second`.

jeaye (changes requested):
Thanks, Shantanu!

test/clojure/core_test/key.cljc
shantanu-sardesai: Done.

jeaye (approved):
Thanks!

PR #951, opened 2026-08-17
Update tests and bump versions for CLJR

keychera:
Update the following cljr tests that have reached some parity with clojure jvm and bump cljr version
- transient
- num

however, I found this one difference between clj and cljr

Clojure JVM
```
Clojure 1.12.3
user=> (num nil)
nil
```

ClojureCLR
```sh
Clojure 1.12.3-alpha8
user=> (num nil)
Execution error (InvalidCastException) at user/eval3824 (NO_FILE:0).
Argument is not numeric
```

I removed that test temporarily with comment

cc: @dmiller for your information

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,7 +134,7 @@ jobs:
 
       - name: Install ClojureCLR
         run: |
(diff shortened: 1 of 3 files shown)


jeaye (approved):
Thanks!

PR #952, opened 2026-08-18
Add test `mod` at big number literals

keychera:
from [findings](https://clojurians.slack.com/archives/C03SRH97FDK/p1786857119903449) on slack

According to the [The Reader](https://clojure.org/reference/reader#:~:text=Integers%20can%20be%20indefinitely%20long) docs, Integers can be indefinitely long so I add test around boundaries of max int and max long.

tested on jvm, cljs, bb, and clr


jeaye (approved):
Thank you!

